PR #2, opened 2020-09-11
Updated Definition Headers

SofiaBrittoSchwartz:
Converted headers (Definition => Java Definition, Declarative Definition => YAML Definition)


reference/refactoring-java-source-code/README.md
```diff
@@ -1,178 +0,0 @@
-# Java Refactoring
```
sambsnyd, line 1: Was this deleted intentionally? If not, bring it back. Same for changetype

sambsnyd (changes requested):

PR #17, opened 2020-12-01
polish checkstyle yaml example

aegershman:
Includes a colon `:` in the checkstyle rewrite yaml example to make the example a lil' more copy-paste-able 👍


sambsnyd:
Thanks! Good polish.

PR #50, opened 2021-05-11
show configuration for having dryRun run on check

aegershman:


reference/gradle-plugin-configuration.md
```diff
@@ -146,6 +146,14 @@ rewrite {
 }
 ```
 
+If desired, `rewriteDryRun` can be configured so that when `check` runs, `rewriteDryRun` does too:
+
+```gradle
```
sambsnyd, line 151: I don't think "gradle" is an option GitBook markdown supports. "groovy" is probably the best fit.

reference/gradle-plugin-configuration.md
```diff
@@ -146,6 +146,14 @@ rewrite {
...
+tasks.named("check").configure {
+    dependsOn(tasks.named("rewriteDryRun"))
+}
```
sambsnyd, line 154: Also include a code snippet that shows how to make dryRun fail if it produces any output, and describe how this can be used to create a gate in CI. I can help you come up with such a snippet if you're not sure how.

sambsnyd (commented):
It would be great to have similar instructions for the maven plugin. That might not be a documentation-only change since Maven is a bit more rigid about these things.

reference/gradle-plugin-configuration.md
aegershman: There's a snippet included right just above this with a snippet about `failOnDryRunResults = true`, if that's what's intended? https://github.com/openrewrite/rewrite-docs/pull/50/files#diff-2825925773c74bc72ef106a6ca7cf732c94644735aacadf104a4324ea0f8be73R140-R146

Can rework the snippet, or have the snippet show both being used together if that's better?

reference/rewrite-maven-plugin.md
```diff
@@ -101,6 +101,31 @@ This goal also produces a report, in the form of a patch file, in which you can
...
+              <goal>dryRun</goal>
+            </goals>
+            <phase>verify</phase>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
+```
```
aegershman: @sambsnyd included demo on how to wire this up in maven as well

PR #62, opened 2021-07-06
doc : template method has been removed

McKratt:
With version 7.8, the  incubating method `JavaVisitor.template()` has been removed, so the doc is no more up to date.


tkvangorder:
Thanks for your contribution, we will include this and also update some other areas that have drifted due to API changes.

PR #64, opened 2021-08-19
fixed wrong parameters on ChangePackage Recipe

Shaoranlaos:
In the example in the quickstart guide the wrong parameter names are used.
These are (no longer?) containing "FullyQualified"

Got the folllwoing errors:
```
[ERROR] Recipe validation error in newPackageName: is required
[ERROR] Recipe validation error in oldPackageName: is required
[ERROR] Recipe validation errors detected as part of one or more activeRecipe(s). Execution will continue regardless.
```

### DIFF
--- a/getting-started/getting-started.md
+++ b/getting-started/getting-started.md
@@ -138,8 +138,8 @@ type: specs.openrewrite.org/v1beta/recipe
 name: com.yourorg.VetToVeterinary
 recipeList:
   - org.openrewrite.java.ChangePackage:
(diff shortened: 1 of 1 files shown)


aegershman:
Just for reference, linking the parameter names in the recipe itself: https://github.com/openrewrite/rewrite/blob/9ec92d5e01e42dab3d40bb3e1bf91a1d1b302f1c/rewrite-java/src/main/java/org/openrewrite/java/ChangePackage.java#L51-L56

Good catch, thank you 👍

PR #76, opened 2022-02-19
Fix example for remove annotation recipe, quote @ in rewrite.yml

fupgang:
Not quoting reserved char @ in rewrite.yml results in
```
[ERROR] found character '@' that cannot start any token. (Do not use @ for indentation)
[ERROR]  in 'reader', line 7, column 26:
[ERROR]           annotationPattern: @java.lang.SuppressWarnings("dep ... 
```


jkschneider:
Thank you @fupgang !

PR #80, opened 2022-04-16
Restore broken-references

timtebeek:
There appear to be [a few broken references](https://github.com/openrewrite/rewrite-docs/search?q=broken-reference).


timtebeek:
@tkvangorder Could you have a look? These broken references appear to have been introduced in 372a66b20e19118a3fc5865bb81e70ae14bc5e0d after 347ffdb671c1fbf904da37292a61d5243dcd4704.

timtebeek:
And perhaps a minimal GitHub Action is in order that fails the build if "broken-reference" is present, especially since a @gitbook-bot commit introduced these issues.

tkvangorder:
yeah, this has been an ongoing issue, when we update the release numbers via a primitive search and replace. Thanks for fixing these. This should get better in the next release because we will be moving to a bom, so only one version will need to be replaced.

PR #88, opened 2022-08-28
Fix direct invocation hints

timtebeek:
Through `grep -rl activeRecipesorg . | xargs sed -i 's/activeRecipesorg/activeRecipes=org/g'`

Could be that these are generated elsewhere; in that case it should be fixed over there.


tkvangorder (approved):
Thanks for the fixing this!

tkvangorder:
We do actually have a recipe document generator, I have fixed this issue there as well. https://github.com/openrewrite/rewrite-recipe-markdown-generator/commit/a43d3fcbf0f55a272fa19c6b8022be908431dcd5

Thanks for all your contributions!

timtebeek:
Figured there must have been a generator. Good to see it fixed there as well, plus now I know where to look. :)

PR #89, opened 2022-08-31
When running recipes manually exceptions can be swallowed

undefz:
Currently in an example for manually running recipes errors are logged only when parsing the source code but not when running recipes.
This may confuse people who are trying to start running recipes, I suggest to use context with logging for both cases.


tkvangorder (approved):
Thank you, this is a good suggestion

PR #92, opened 2022-10-06
Fixed closing <scope> tags

charlesfinley:
Some scope tags were missing the closing forward slash.


pway99 (approved):
Thanks @charlesfinley

PR #97, opened 2022-11-21
Update the getting started guide

mike-solomon:
* Made the introduction clearer
* Added a prerequisites section so people can see what you need to know
* Provided more context in the steps
* Corrected various bugs in the files
* Ensured style for the files was consistent with the actual code
* Broke up the Maven and Gradle external modules section so it's more clear

If any of you have any suggestions for how to make things clearer or if you think something sounds weird or wrong please let me know :)

### DIFF
--- a/getting-started/getting-started.md
+++ b/getting-started/getting-started.md
@@ -88,10 +88,6 @@ rewrite {
 // ...
 ```
 {% endcode %}
-
-{% hint style="info" %}
-When applied to a multi-project build, behavior differs depending on whether the plugin is applied to the root project or to a sub-project. Applied to the root project, the plugin will parse and refactor all sources from all projects. Applied to any project other than the root project, the plugin will parse and refactor only sources from that project.
-{% endhint %}
 {% endtab %}
 {% endtabs %}
 
--- a/getting-started/getting-started.md
+++ b/getting-started/getting-started.md
@@ -27,7 +27,7 @@ This quick start guide assumes that you:
 
 ## Step 1: Clone sample project
 
-The first step in this process is making sure you have some code to work with. We've created a [sample repository](https://github.com/openrewrite/spring-petclinic-migration) that you can use if you'd like. However, as all of the steps in this guide apply to any Java project built with Maven or Gradle, please feel free to use your own and skip to [Step 2](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project).
+The first step in this process is making sure you have some code to work with. We've prepared a [sample repository](https://github.com/openrewrite/spring-petclinic-migration) that you can use if you'd like. However, as all of the steps in this guide apply to any Java project built with Maven or Gradle, please feel free to use your own and skip to [Step 2](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project).
 
 {% hint style="warning" %}
 The sample spring-petclinic project requires JDK version 11 to build. Newer JDK versions will not work. Get OpenJDK 11 [here](https://adoptium.net/temurin/releases/?version=11) if you do not already have that version installed.
--- a/getting-started/getting-started.md
+++ b/getting-started/getting-started.md
@@ -1,39 +1,53 @@
 ---
 description: >-
-  This quick-start guide covers configuring rewrite with Maven and Gradle and
-  using it to apply refactoring recipes.
+  This guide covers configuring rewrite with Maven or Gradle and using it to run refactoring recipes.
 ---
 
-# Quickstart: Maven and Gradle
+# Quick Start: Setting Up Your Project and Running Recipes
 
-Following this guide you will learn:
+To help orient you to OpenRewrite, let's walk through configuring a project to use the Maven or Gradle rewrite plugin. Then let's walk through running various types of recipes on said project and talk through the results.
 
-* How to apply the rewrite-maven-plugin to a Maven project
-* How to apply the rewrite-gradle-plugin to a Gradle project
-* How to configure, run, and inspect the results of built-in and external refactoring recipes
+In this guide you will:
+
+* [Clone a sample project](#step-1-clone-sample-project)
+* [Add the rewrite-maven-plugin or rewrite-gradle-plugin to your project](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project)
+* [Activate a recipe so it can be run](#step-3-activate-a-recipe)
+* [Run a simple refactoring recipe](#step-4-run-a-simple-refactoring-recipe)
+* [Run a recipe with YAML configuration](#step-5-run-a-recipe-with-yaml-configuration)
+* [Add and run an externally created recipe](#step-6-running-recipes-from-external-modules)
+
+## Prerequisites
+
+This quick start guide assumes that you:
+
+* Are somewhat familiar with Java
+* Have worked with a Maven or a Gradle project before
+* Know how to run commands in the command line
+* Can run basic Git commands
 
 ## Step 1: Clone sample project
 
+The first step in this process is making sure you have some code to work with. We've created a [sample repository](https://github.com/openrewrite/spring-petclinic-migration) that you can use if you'd like. However, as all of the steps in this guide apply to any Java project built with Maven or Gradle, please feel free to use your own and skip to [Step 2](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project).
+
 {% hint style="warning" %}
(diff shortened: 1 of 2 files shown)


getting-started/getting-started.md
```diff
@@ -62,6 +84,8 @@ repositories {
 rewrite {
     // Will configure in subsequent steps
 }
+
+// ...
 ```
 {% endcode %}
```
mike-solomon, line 91: I didn't follow the hint below when I was reading through this guide but I left it in as I presumed it was important. If one of y'all could take a look and double-check (and perhaps provide me with some additional context) that would be swell!

```When applied to a multi-project build, behavior differs depending on whether the plugin is applied to the root project or to a sub-project. Applied to the root project, the plugin will parse and refactor all sources from all projects. Applied to any project other than the root project, the plugin will parse and refactor only sources from that project.```

File: getting-started/getting-started.md [hunk 2]
```diff
@@ -185,39 +224,55 @@ This creates a new recipe called `com.yourorg.VetToVeterinary`. Now add it to th
...
+
+Below, we'll walk through the [Maven](#maven--external-modules) and [Gradle](#gradle--external-modules) changes and provide some additional context around said changes.
+
+### Maven + External Modules
+
+For Maven projects, you'll need to: 
+
+* Add the recipe to the `activeRecipes` list
+* Add a dependency on `rewrite-spring` (where the JUnit 4 to 5 recipe lives)
+* Specify a version of `rewrite-spring` to use
```
mike-solomon, line 272: I'm unsure how someone would know which version to specify for a given recipe. It seems like there's a possibility that people could specify the wrong version or have versions that are mismatched. Is that true? Could anyone give me some additional context on that front?

getting-started/getting-started.md
tkvangorder: I think this is really the expected behavior (of any gradle build task): 

if you run `./gradlew rewriteDryRun`, I think most developers would expect that to work on the entire project, whereas, `./gradlew :subProject:rewriteDryRun` would just work on the subproject. 

Not sure if there is something specific we are trying to call out here. I might ask @sambsnyd

getting-started/getting-started.md
tkvangorder: Well, we actually specify the version in the example pom.xml. We update that version with each release.

getting-started/getting-started.md
```diff
@@ -1,39 +1,53 @@
...
+This quick start guide assumes that you:
+
+* Are somewhat familiar with Java
+* Have worked with a Maven or a Gradle project before
+* Know how to run commands in the command line
+* Can run basic Git commands
 
 ## Step 1: Clone sample project
 
+The first step in this process is making sure you have some code to work with. We've created a [sample repository](https://github.com/openrewrite/spring-petclinic-migration) that you can use if you'd like. However, as all of the steps in this guide apply to any Java project built with Maven or Gradle, please feel free to use your own and skip to [Step 2](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project).
```
sambsnyd, line 30:
```suggestion
The first step in this process is making sure you have some code to work with. We've prepared a [sample repository](https://github.com/openrewrite/spring-petclinic-migration) that you can use if you'd like. However, as all of the steps in this guide apply to any Java project built with Maven or Gradle, please feel free to use your own and skip to [Step 2](#step-2-add-rewrite-maven-plugin-or-rewrite-gradle-plugin-to-your-project).
```
I'm changing the word "created" to "prepared", because spring-petclinic was created by the spring team as a demo spring app. We forked that and made it suitable for our purposes.

getting-started/getting-started.md
sambsnyd: That note could probably be mentioned only in the documentation on the gradle plugin itself, rather than here. Someone following our prescriptive guidance would not have to make any decision around this

getting-started/getting-started.md
```diff
@@ -185,39 +224,55 @@ This creates a new recipe called `com.yourorg.VetToVeterinary`. Now add it to th
...
+        'com.yourorg.VetToVeterinary'
+    )
+}
 ```
 {% endcode %}
 {% endtab %}
 {% endtabs %}
 
-Run `./mvnw rewrite:run` or `./gradlew rewriteRun` to perform the package relocation. Afterward you'll see both that the sources in the vet package have been moved to the new directory, and references such as import statements have been updated accordingly.
+Once this recipe has been added to your active recipes, you can run either `./mvnw rewrite:run` or `./gradelw rewriteRun` to execute all of your active recipes. Afterward, you'll see that:
```
sambsnyd, line 244:
```suggestion
Once this recipe has been added to your active recipes, you can run either `./mvnw rewrite:run` or `./gradlew rewriteRun` to execute all of your active recipes. Afterward, you'll see that:
```

getting-started/getting-started.md
```diff
@@ -46,11 +60,19 @@ In the pom.xml or build.gradle, add this entry to the `plugins` section to apply
...
+Your file should look similar to:
+
 {% code title="build.gradle" %}
 ```groovy
 plugins {
-    id("java")
-    id("org.openrewrite.rewrite") version("5.32.0")
+    id 'java'
+    id 'maven-publish'
+    id 'org.openrewrite.rewrite' version '5.32.0'
```
sambsnyd, line 75: The changes you've made to omit parenthesis and prefer single-quoted to double-quoted strings _do_ make these more idiomatic groovy. But Gradle scripts can be written in Groovy or Kotlin. And Kotlin has no single-quoted string and does not allow parenthesis to be omitted in these circumstances.

But part of why I used double-quoted strings and included parenthesis is that the resulting code would, generally, work _either_ in groovy or kotlin scripts. If we're going to make the groovy in these code snippets more idiomatic, which is obviously not a bad thing, we might consider _also_ including gradle kotlin script examples.

getting-started/getting-started.md
```diff
@@ -245,43 +300,58 @@ After applying these steps, the relevant portions of your build file will look l
...
-{% endtab %}
 
-{% tab title="Gradle" %}
+To double-check that everything is working, run the command `./mvnw rewrite:run`. Your project should be upgraded to Spring Boot 2 and all of the test classes should be updated to JUnit 5. Your `pom.xml` file will also have had its Spring dependencies updated, the JUnit 4 dependency removed, and the JUnit 5 dependency added.
+
+### Gradle + External Modules
+
+Unlike Maven projects, Gradle projects have two options for specifying versions for recipes. You can:
+
+1. Add `rewrite-recipe-bom` as a dependency
```
sambsnyd, line 310:
```suggestion
1. Add `rewrite-recipe-bom` as a [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import)
```

Adding rewrite-recipe-bom as a regular dependency is not the same as using it to manage the versions of other dependencies. You accurately explain the difference below, but I don't want a discrepancy to cause confusion.

getting-started/getting-started.md
```diff
@@ -245,43 +300,58 @@ After applying these steps, the relevant portions of your build file will look l
...
+2. Add `rewrite-spring` as a dependency with a version specified
+
+If you choose to use the `rewrite-recipe-bom`, you won't have to worry about specifying versions for your rewrite recipes as all of the recipes you include in your `dependencies` section will have an appropriate version specified in the bill of materials (BOM). For Gradle projects, this is the recommended approach.
+
+If you choose to not use the `rewrite-recipe-bom`, you'll need to ensure that you have the appropriate version of all of your dependencies included (as you would with Maven projects).
```
sambsnyd, line 315:
```suggestion
If you choose to not use `rewrite-recipe-bom`, you'll need to specify the version of each rewrite recipe module you use. 
```

mike-solomon: That's a great call out. Agree with this suggestion. Thanks

getting-started/getting-started.md
mike-solomon: That's fair re the Kotlin side of things. The reason I switched this, though, is because when I checked out the sample repository that we suggest, it doesn't use the double quotes and parenthesis. I figure if someone was following along, it would be nicer to have the file look like it would in the sample repository. 

Would you mind explaining what you mean by `also including gradle kotlin script examples`? I'm not sure I follow that.

getting-started/getting-started.md
mike-solomon: Good call out and appreciate the suggested change.

getting-started/getting-started.md
mike-solomon: Alrighty -- I removed it to reduce confusion. Thanks!

getting-started/getting-started.md
mike-solomon: Going to merge this for now but happy to update later once I better understand the ask :)

PR #100, opened 2022-12-05
Add an AST Element Example Doc

mike-solomon:
The purpose of this doc is to show people what code corresponds with which AST element. It also explains how to use the debugger to further dig into these elements.

Would appreciate someone looking over the terminology used in this doc. I'm unsure if the way I phrased things make sense or if the names of certain elements are correct. Would also be good to double-check that the appropriate things are highlighted in the diagrams.


reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
...
+        return a;
+    }
+}
+
+@interface AnAnnotation {}
+```
+
+### Binary
+
+A [Binary element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L597-L705) is any operation with a left and right side (usually separated by something like a `+`).
```
sambsnyd, line 52:
```suggestion
A [Binary](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L597-L705) is an [Expression](#expression) with a left and right side, separated by an operator. Examples of operators include `+`, `-`, `||`, `&&`, and more.
```

reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
...
+
+![Binary Element Example](<../.gitbook/assets/Binary.png>)
+
+### Block 
+
+A [Block element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L712-L851) is the code contained between curly braces. Blocks can be nested inside of each other.
```
sambsnyd, line 58:
```suggestion
A [Block](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L712-L851) is a pair of curly braces and the [Statements](#statement) contained within. Blocks can be nested inside of each other.
```

reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
+# AST Element Examples
+
+When building recipes, it's important to understand how certain elements in the OpenRewrite Abstract Syntax Tree (AST) correspond to code. You couldn't, for example, properly rename a variable with a recipe unless you knew that `J.Identifier` is how variable names are represented in the AST.
```
sambsnyd, line 3: I note the term "element" is used extensively throughout this document, but "element" isn't a term we define anywhere.

I wonder if using this term repeatedly may be confusing. It feels like it implies some formal distinction or concept that isn't there. "not complete/total" or "component of" is what I usually mean by "element" when I say "AST element", but readers might be wondering if it means something more formal. Is a `J.CompilationUnit` an "element" of an AST since it is a whole AST?

reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
...
+
+### FieldAccess
+
+A [FieldAccess element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L1827-L1953) is any fully qualified name. Often times, these are package or import statements, but they can also appear in code as something like: `this.foo`.
+
+![FieldAccess Element Example](<../.gitbook/assets/FieldAccess.png>)
+
+### Identifier
+
+An [Identifier element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L2310-L2343) is any name in the code (class names, variable names, method names, etc).
```
sambsnyd, line 90: Good place to mention that `J.Identifier.getFieldType()` can be used to tell what class the identifier is a field on. `null` for identifiers which are not fields.

reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
...
+
+![Identifier Element Example](<../.gitbook/assets/Identifier.png>)
+
+### MethodDeclaration
+
+A [MethodDeclaration element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L3517) is all the code for creating a method.
```
sambsnyd, line 96:
```suggestion
A [MethodDeclaration](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L3517) is the annotations, modifiers, return type, name, argument list, and body which together define a method on a [class](#classdeclaration).
```

sambsnyd (changes requested):

reference/ast-examples.md
sambsnyd: Part of the language trickiness here is than an AST element is itself an AST. So in conversation we normally say "AST" to mean "the whole AST" and "AST element" to mean "any AST which is itself a component of another, larger AST"

tkvangorder: This is an interesting distinction. I think I have been guilty in the past of calling them "elements of an AST". Are you suggesting replacing `element` with `ast`? I think this makes sense the more I think about it...

"AST Element Examples" could just be "AST Examples"..."under how certain ASTs in the ....." etc.

reference/ast-examples.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Block 
+
+A [Block element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L712-L851) is the code contained between curly braces. Blocks can be nested inside of each other.
+
+![Block Element Example](<../.gitbook/assets/Block.png>)
+
+### ClassDeclaration
+
+A [ClassDeclaration element](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L1062-L1336) contains all of the code for any Java class.
```
tkvangorder, line 64: I think it would be a good idea to remind users that a ClassDeclaration can be nested inside another class: 

```
public class A {
...

   private class B {
   ...
   }
}
```

tkvangorder (changes requested):
@mike-solomon This looks really good and I love the visualizations. This is going to be very helpful to our users.

reference/ast-examples.md
```diff
@@ -0,0 +1,185 @@
...
+
+For instance, you could have a [CompilationUnit](#compilationunit) which then has a [ClassDeclaration](#classdeclaration) inside of that which then has a [MethodDeclaration](#methoddeclaration) inside of that and so on. On top of being a valid AST that would _also_ represent valid Java code. 
+
+On the other hand, you _could_ define a [Binary](#binary) with other Binaries inside of that. That would be a valid AST but it **would not** represent valid Java code as just entering `1+2` (a valid Binary) in a Java file would not compile (that code would need to be in an expression which is then in a class to be valid).
+
+> Every object in the [code section](#code) is an AST in the sense that it implements a Tree interface. These objects, however, need to be combined in an appropriate way to represent valid Java. It is your duty as a recipe author to ensure that the final AST represents valid Java code.
```
sambsnyd, line 32:
```suggestion
For instance, a [ClassDeclaration](#classdeclaration) is an AST which defines a class. A typical class declaration will be composed of elements such as fields, methods, constructors, and inner classes. Each of these are represented as their own, smaller AST. So the term "AST" may refer to an entire, complete Java file or just one piece of it.

It is possible to manipulate ASTs to create code which will not compile. OpenRewrite provides some safeguards against grammatically invalid transformations in its type system - you can't replace an import statement with a method declaration, for example. But there are plenty ways ASTs can be valid according to the Java grammar without being valid, compilable programs. For example, nothing prevents you from modifying an AST such that a variable is used before it is defined. It is the responsibility of recipe authors to consider language semantics and the full range of possibilities when making changes. In accordance with the principle of Doing No Harm, always err on the side of leaving code untouched rather than making a risky change. 
```

Readers don't yet know what a `CompilationUnit` or `ClassDeclaration` are, so these terms cannot yet be used to explain related concepts.
A java programmer is likely to infer the meaning of `ClassDeclaration` correctly, but I had no idea that the name for an entire Java file was a "compilation unit" until I worked here.

reference/ast-examples.md
```diff
@@ -0,0 +1,185 @@
...
+        return a;
+    }
+}
+
+@interface AnAnnotation {}
+```
+
+### Binary
+
+A [Binary AST](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L597-L705) is an [Expression](#expression) with a left and right side, separated by an operator. Examples of operators include `+`, `-`, `||`, `&&`, and more.
```
sambsnyd, line 64: The line numbers in these links will be wrong as soon as we make changes to `J`. Which aren't constant, but it does happen. Just today I tweaked an entry in `J`. If we can't make more durable links, and I'm not aware of any way to do that, better to link to `J` once without trying to link to its individual components.

reference/ast-examples.md
```diff
@@ -0,0 +1,185 @@
+# AST Examples
+
+When building recipes, it's important to understand how OpenRewrite Abstract Syntax Trees (AST) correspond to code. You couldn't, for example, properly rename a variable with a recipe unless you knew that [J.Identifier](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L2310-L2343) is the class used to represent a variable.
+
+To help you get started on working with ASTs and OpenRewrite, this guide will:
```
sambsnyd, line 5: We should mention that this is a guide to the Java AST. XML, Protobuf, Yaml, etc., each have their own AST. Being data formats those are generally simpler and more approachable structures.

sambsnyd (changes requested):
Getting closer. Left some suggestions which I believe will help clarify things

reference/ast-examples.md
```diff
@@ -0,0 +1,185 @@
...
+        // ...
+    }
+}
+```
+
+![ClassDeclaration Example](<../.gitbook/assets/ClassDeclaration.png>)
+
+### CompilationUnit
+
+A [CompilationUnit AST](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L1342-L1545) is the root of the Java AST. In order for an AST to represent valid Java code, all other elements must be contained inside of this.
```
sambsnyd, line 91: I don't think we need to say "<AST type> AST" for each of these. It is redundant. 

It's like if we were listing breeds of dogs like this:

- Labrador dog
- Schnauzer dog
- Beagle dog
- Dachshund dog
- Corgi dog

Probably didn't need so many dog types to illustrate my point. But look how cute this beagle is:
image

reference/ast-examples.md
mike-solomon: I could update the links to use a tag such as: https://github.com/openrewrite/rewrite/blob/v7.33.0/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java which will stay consistent regardless of the changes made. 

Does that seem reasonable to you or would you prefer I just link to the file rather than the individual parts?

reference/ast-examples.md
mike-solomon: I'm okay with making this change -- but I do want to call out that I'm not sure I agree with that. The naming conventions we have don't make it obvious that these things _are_ ASTs. If I saw a section called, `Identifier` and saw that it talked about variable names, I definitely would not assume it's an AST. That being said, I recognize that I'm still learning this so will default to you on this.

reference/ast-examples.md
```diff
@@ -0,0 +1,183 @@
+# AST Examples
+
+When building recipes, it's important to understand how OpenRewrite Abstract Syntax Trees (AST) correspond to code. You couldn't, for example, properly rename a variable with a recipe unless you knew that [J.Identifier](https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/tree/J.java#L2310-L2343) is the class used to represent a variable.
```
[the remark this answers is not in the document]
knutwannheden: Perfect! I missed that comment.

PR #103, opened 2022-12-12
Update recipe development environment getting started guide

mike-solomon:
* Removes references to Kotlin that were no longer applicable
* Adds comments to confusing sections of code
* Adds missing Java dependency
* Cleans up project layout to be more accurate
* Provides a better explanation for local testing
* Minor grammar / spelling fixes

### DIFF
--- a/getting-started/recipe-development-environment.md
+++ b/getting-started/recipe-development-environment.md
@@ -11,9 +11,9 @@ This getting started guide covers setting up your development environment for cr
 * [JDK](https://adoptopenjdk.net) (version 1.8+)
   * A JRE alone is insufficient since OpenRewrite uses compiler internals and tools only found in the JDK
 * [Gradle](https://gradle.org) (version 4.0+ ) or [Maven](https://maven.apache.org) (version 3.2+)
-* Text Editor or IDE with Java support. [Kotlin](https://kotlinlang.org) support optional but recommended for a good test authoring experience
+* Text Editor or IDE with Java support
   * [IntelliJ](https://www.jetbrains.com/idea/download/)
-  * [Eclipse](https://www.eclipse.org/downloads/) with [Kotlin Plugin](https://marketplace.eclipse.org/content/kotlin-plugin-eclipse)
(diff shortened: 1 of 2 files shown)


getting-started/recipe-development-environment.md
```diff
@@ -111,6 +115,11 @@ dependencies {
         <artifactId>rewrite-java-11</artifactId>
         <scope>runtime</scope>
     </dependency>
+    <dependency>
+        <groupId>org.openrewrite</groupId>
+        <artifactId>rewrite-java-17</artifactId>
+        <scope>runtime</scope>
+    </dependency>
```
mike-solomon, line 122: I'm pretty sure we just forgot to add a `java 17` section here. Let me know if I misunderstood something, though.

getting-started/recipe-development-environment.md
```diff
@@ -54,6 +54,8 @@ dependencies {
 
     // rewrite-java dependencies only necessary for Java Recipe development
     implementation("org.openrewrite:rewrite-java")
+
+    // add based on your Java needs
```
mike-solomon, line 58: I wanted to clarify that you don't need all of these for a given project.

sambsnyd: Technically all that's needed is the one corresponding to the JDK version you'll be running on. They can coexist on a classpath. So it's a bit redundant to have all three, but it's also maximally flexible for whatever JDK version they end up using.

sambsnyd: Technically all that's needed is the one corresponding to the JDK version you'll be running on. They can coexist on a classpath. So it's a bit redundant to have all three, but it's also maximally flexible for whatever JDK version they end up using. There's no need for users to spend effort considering this point if they accept the default. If we're going to have a comment explaining, which is fine, it should probably not imply that they need to consider and make a choice when no such consideration or choice is necessary.

sambsnyd (changes requested):
Looks good overall

mike-solomon:
Merging in the changes. Please feel free to let me know if I should change anything else or if the changes I made were incorrect :)

PR #105, opened 2022-12-16
Rewrite the Java Refactoring Recipe tutorial

mike-solomon:
This addresses a lot of issues with this tutorial such as:
* Many parts of the code were wrong or unneeded
* Kotlin was being used and encouraged when it shouldn't be
* There was not enough context for people to understand certain topics
* Some hints or comments were incorrect

### DIFF
--- a/tutorials/authoring-recipes/writing-a-java-refactoring-recipe.md
+++ b/tutorials/authoring-recipes/writing-a-java-refactoring-recipe.md
@@ -4,12 +4,13 @@ description: Adding a method to a class that returns a String
 
 # Writing a Java Refactoring Recipe
 
-In this tutorial, we'll create a basic refactoring recipe that adds a method returning a `String` to a user-specified class. This `SayHelloRecipe` will take a class like this:
+To help you get started with writing recipes, this guide will walk you through all the steps needed to create a basic refactoring recipe. This
+`SayHelloRecipe` will add a `hello()` method to a user specified class if that class does not already have one. For example, it would take a class like this:
 
 ```java
 package com.yourorg;
 
-class A {
+class FooBar {
 
 }
 ```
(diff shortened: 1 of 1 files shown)


tutorials/authoring-recipes/writing-a-java-refactoring-recipe.md
```diff
@@ -19,46 +20,51 @@ And refactor it into a class like this:
...
             example = "com.yourorg.FooBar")
     @NonNull
-    private final String fullyQualifiedClassName;
-    public String getFullyQualifiedClassName() {
-        return fullyQualifiedClassName;
-    }
+    String fullyQualifiedClassName;
 
-    // Recipes must be serializable. This is verified by RecipeTest.assertChanged() and RecipeTest.assertUnchanged()
+    // All recipes must be serializable. This is verified by RewriteTest.rewriteRun() in your tests.
```
tkvangorder, line 67: Pretty sure you can remove this manually created constructor, it will be generated via Lombok's `@Value` annotation. You do not need the `JsonProperty` annotation if you add `-parameters` as a compiler option.

tkvangorder: I think it would be enough to say that "A recipe's state is serialized to/from JSON using the Jackson serialization library. The serializability of a recipe is verified by RewriteTest.rewriteRun() in your tests."

tkvangorder (approved):
Looks great Mike!

tutorials/authoring-recipes/writing-a-java-refactoring-recipe.md
mike-solomon: @tkvangorder Thanks for the call out! I'm not sure what you mean by `-parameters` as a compiler option, though. Would you mind expanding on what people would need to do for that? I saw that my tests ran just fine if I removed the constructor as you suggested so I'm happy to update the docs and remove it altogether. Unsure if there's a concern with that, though. Maybe I'll add the serialize comment to the comment above the lombok annotation?

PR #107, opened 2022-12-22
Rename Abstract Syntax Tree to Lossless Semantic Tree

mike-solomon:
* Attempts to update all docs to utilize LST instead of AST.
* Also minor rephrases in the LST explanation doc.
* Also minor grammar / spelling fixes.


concepts-and-explanations/lossless-semantic-trees.md
```diff
@@ -0,0 +1,21 @@
+# Lossless Semantic Trees
+
+A Lossless Semantic Tree (LST) is a tree representation of code. Unlike the traditional [Abstract Syntax Tree](https://en.wikipedia.org/wiki/Abstract\_syntax\_tree) (AST), Moderne's LST offers a unique set of characteristics that make it suitable for both single-repository and mass transformation operations:
+
+* **Type-attributed**. Each LST element is imbued with type information. For example, when referencing a field, the source code may just refer to it as `myField`. The OpenRewrite LST element for `myField`, on the other hand, would contain additional information about what the type of `myField` is, even if it isn't defined in the same source file or even the same project.
```
traceyyoshima, line 5: Curious if you think it's helpful to include `each type-attributed LST?` since the term lossless is contextual to the language.

Mainly to acknowledge there are differences between types and non-typed languages like `XML` and `YML` compared to `Java`, and `Groovy`.

mike-solomon: That's a fair point! I do think it's useful to point out that there is knowledge beyond the text for languages that have types, but I can see your point about it being part of the "lossless" definition and not applicable to non-typed languages.

If you have any suggestions for phrasing, would love to hear them. If not, I can ponder more about it. 

Thanks for the input :D

tkvangorder: I would change "Moderne's LST" to "OpenRewrite's LST"

tkvangorder: Honestly, I view "lossless" to mean..."maintain all formatting" and "semantic" to mean type attributed.

mike-solomon: @tkvangorder - I think that's reasonable to think, but I also believe we should call these things out. "Semantic" is a term that many people find especially difficult to understand.

traceyyoshima: I agree with Mike since semantics is, by definition something that may differ from person to person.

We frequently interact with engineers from different cultures, who speak different languages, and are at different points in their careers.

I am unsure how we communicate the concepts to the reader, but I think it is essential to be clear so that our readers understand the framework and how to use it.

mike-solomon:
Going to merge this in as I haven't heard any other feedback. If there's suggestions later, please feel free to let me know!

PR #117, opened 2023-01-16
Update plugin version references

timtebeek:
Hope this doesn't conflict with any changes you hadn't pushed yet; it's simple search replace on plugin version numbers.


mike-solomon:
LGTM. As a gentle reminder, though, all docs in `/reference/recipes` are auto-generated so any changes you made here will be clobbered when we re-generate the docs. See https://github.com/openrewrite/rewrite-recipe-markdown-generator for where that's done.

timtebeek:
That's OK; this was mostly to fix the versions used in the guides, as I saw an issue using an incorrect version.

PR #123, opened 2023-01-30
Add multiple visitors guide

mike-solomon:
This adds a tutorial for making a reasonably complex recipe with multiple visitors.

Would appreciate any feedback on if the suggested approaches make sense or if there's anything incorrect or confusing. Thanks :)


authoring-recipes/multiple-visitors.md
```diff
@@ -0,0 +1,814 @@
...
+        public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations varDec, ExecutionContext executionContext) {
+            J.VariableDeclarations variableDeclarations = super.visitVariableDeclarations(varDec, executionContext);
+            
+            // If the varDec already has a "final" modifier, we don't need to check anything else
+            if (variableDeclarations.hasModifier(J.Modifier.Type.Final)) {
+                return variableDeclarations;
+            }
+
+            // Uninitialized local variables should not be considered
+            if (variableDeclarations.getVariables().stream().anyMatch(namedVariable -> namedVariable.getInitializer() == null)) {
```
knutwannheden, line 426: This is not quite true and our recipe could be improved to finalize more local variables. If the compiler can determine that a final variable is assigned exactly once before it is first dereferenced, it doesn't have to be assigned using an initializer. So the following is also perfectly valid:
```java
final int x;
if (...) {
  x = 3;
} else {
  x = 4;
}
```
We could enhance our recipe to also add the `final` modifier in these cases. But I am wondering if it is worth the trouble.

I just wanted to point this out, in case we want to change the comment to point out that we intentionally ignore these more complex cases.

mike-solomon: Thanks for the feedback @knutwannheden . I know it's possible to expand on this recipe to handle more cases (and I think @kunli2 is working on that). However, I wanted to limit the scope of the recipe to mostly what exists in it right now. 

In the scope section at the beginning of the document, I call out that for the purposes of this guide, we won't consider uninitialized variables: https://github.com/openrewrite/rewrite-docs/blob/8fa687ce1024816ab34fe9bf69532a4274f729cd/authoring-recipes/multiple-visitors.md?plain=1#L42

Do you think it would help if I re-iterate that here? Like change the comment to be `// Uninitialized local variables should not be considered in this tutorial`?

knutwannheden: No, I missed that part in the scope. Thanks for clarifying!

kunli2 (approved):

authoring-recipes/multiple-visitors.md
kunli2: Just FYI, I am going to improve the recipe `FinalizeLocalVariables` to cover more case (include without initializer), but for the case of `if / else` as Knut exampled above.
```java
final int x;
if (...) {
  x = 3;
} else {
  x = 4;
}
```
I haven't got a solution for this yet, so we will leave it uncovered.

PR #128, opened 2023-02-01
Add YAML LST Examples Doc

mike-solomon:


concepts-and-explanations/yaml-lst-examples.md
```diff
@@ -0,0 +1,194 @@
...
+---
+- item_1
+- item_2
+```
+
+## YAML LSTs
+
+### Documents
+
+The [Documents LST](https://github.com/openrewrite/rewrite/blob/v7.35.0/rewrite-yaml/src/main/java/org/openrewrite/yaml/tree/Yaml.java#L65-L129) is the root of the YAML LST. In order for an LST to represent valid YAML, all other elements must be contained inside of this. It is composed of one or more [Document LSTs](#document).
```
mike-solomon, line 52: I struggled a bit with thinking of how to phrase this as `Documents` could mean an object or it could mean many `Document` objects. I decided to go with adding `LST` to the end of these so it was a bit clearer. Happy to change the phrasing if someone else has a strong preference on this (or if this phrasing is wrong for some reason).

concepts-and-explanations/yaml-lst-examples.md
```diff
@@ -0,0 +1,194 @@
...
+* [Discuss how that YAML relates to different types of LSTs](#yaml-lsts)
+* [Teach you how to learn more about YAML LSTs yourself](#using-the-debugger-to-detect-lsts) 
+
+{% hint style="info" %}
+If you want a more thorough explanation of what LSTs are and how they're used in recipes, please read [this high-level LST explanation](/concepts-and-explanations/lst-examples.md#high-level-lst-explanation).
+{% endhint %}
+
+## Sample YAML
+
+Below is a simple YAML file whose entire purpose is to demonstrate different types of LSTs. Each of the following sections will highlight different parts of this file to demonstrate which chunks correspond to which LST. This listing of LSTs is not exhausting, but it should give you a good sense of the most common types.
```
timtebeek, line 17:
```suggestion
Below is a simple YAML file whose entire purpose is to demonstrate different types of LSTs. Each of the following sections will highlight different parts of this file to demonstrate which chunks correspond to which LST. This listing of LSTs is not exhaustive, but it should give you a good sense of the most common types.
```
(I could be wrong)

mike-solomon: Good catch - you're right :)

timtebeek (approved):
Looks like a good gentle introduction, well done!

Issue #126, opened 2023-01-31
Document how to consume snapshot versions from the snapshot pages


timtebeek:
Quite like that we now have snapshots listed in the documentation: https://docs.openrewrite.org/changelog/snapshot-2023-01-30

Next question that came to my mind seeing those is: how would I consume those as a user? 
Is there a BOM published to a snapshot repository that I can use? Likely yes, but documentation could help. :)

mike-solomon:
Thanks for the suggestion! 

To document what I think needs to happen:
* Add a Gradle section that:
  * Specifies how to update the `repositories` section to add a `maven` configuration that points to a `url` of "https://oss.sonatype.org/content/repositories/snapshots/" (which contains the snapshots)
  * Specifies how to update their recipe versions (by using `latest.integration` or a specific version). For instance: `(platform("org.openrewrite:rewrite-bom:latest.integration"))`
* Add a Maven section that:
  * Does something similar as the Gradle section (have to figure out what that is). Important to keep in mind that `latest.integration` doesn't work for Maven. 
  
Will need to figure out if we can grab version numbers in the markdown generator to add to this. Also need to figure out where to put this. For instance, should this be added at the bottom of every snapshot changelog? Should it be its own standalone doc? Something else?

mike-solomon:
Updated markdown generator to add a link to this doc: https://github.com/openrewrite/rewrite-recipe-markdown-generator/pull/42

With that this should be all done :)

PR #130, opened 2023-02-02
Add instructions for how to use snapshot versions

mike-solomon:
Right now we don't have any instructions for how to use snapshot versions. I think adding this reference, in combination with linking to it from each snapshot changelog we generate, would be a reasonable solution to: https://github.com/openrewrite/rewrite-docs/issues/126

### DIFF
--- a/reference/snapshot-instructions.md
+++ b/reference/snapshot-instructions.md
@@ -1,0 +1,71 @@
+# How to use snapshot versions
+
+Right now, OpenRewrite does a full release once every 2-4 weeks. Between those releases, recipes are consistently being added or changed to meet the needs of our users. For many people, it's fine to wait until a new release to get access to these new features or recipes. Others, however, would like to have access right away.
+
+This guide will provide you with instructions on how to configure your project to access these "snapshot versions" so you can get updates in real time.
+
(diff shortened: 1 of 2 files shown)


reference/snapshot-instructions.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Gradle instructions
+
+To utilize snapshot versions in Gradle projects, you'll need to make the following changes to your `build.gradle` file:
+
+* Update your `repositories` section to include the location of the OpenRewrite snapshots:
+
+```groovy
+repositories {
+    // ...
```
timtebeek, line 15: I'm wondering if it's sufficiently clear that people could have (or should have) other repositories mentioned as well. I want to prevent users only adding this block in it's entirety, and then not being able to resolve dependencies from Maven central. Other ways to phrase this welcome.

```suggestion
    // ... possible other repositories, such as mavenCentral()
```

timtebeek (approved):

reference/snapshot-instructions.md
mike-solomon: 👍 sounds good Will change.

Issue #124, opened 2023-01-31
Document the new `classPathFromResources` with bundled jars


timtebeek:
As [reported on Slack](https://rewriteoss.slack.com/archives/C01A843MWG5/p1675097259855829) there was a bit of confusion on how (and when) best to use the new `classpathFromResources` to setup a recipe test classpath. So far the best documentation to that is in private Slack threads [one](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1674576044031909) and [two](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1674804578685159).

mike-solomon:
Another useful thread to add to this list: [thread](https://rewriteoss.slack.com/archives/C01A843MWG5/p1675929089785139?thread_ts=1675894219.267039&cid=C01A843MWG5).

mike-solomon:
Want to double-check my understanding of `classPathFromResources`. @sambsnyd @timtebeek - would one of you mind double-checking that the below information is correct? I can then clean this up and make a doc for it.

## The problem

If a recipe needs more than two versions of a library, it is can be confusing on how to set that up so tests can run correctly. This is common with migration recipes that span multiple versions. 

Let's consider a recipe that migrates something from Java 11 to 15, but along the way breaks it up into `Java 11 -> 12` and `Java 12 -> 13` and so on. The recipe itself will produce code that is Java 15 code. The tests, however, might expect code to be Java 11 or 12 or 13 or 14, though. If you specify multiple dependencies on different versions in your build files, only the top one will be used. 

This can cause people to struggle to get the appropriate libraries configured for specific tests. Please note, though, that this is only an issue when the tests need many different versions in them. If the tests only need one version, then using `classPathFromResources` doesn't have much use (?)

## The solution

The rewrite recipe library now offers some new functionality to assist with this. It will look in the `META-INF` folder for dependencies. Many different versions of the same dependency can be included in there. Your code can then specify exact versions to run for a particular test. 

For instance, if you wanted to specify that one test should use `assertj-core-3.24.2`, you could add this to your spec:

```java
spec
    .parser(JavaParser.fromJavaVersion()
        .classpathFromResources("assertj-core-3.24.2")
```

This would look in the `META-INF` folder for an `assertj-core-3.24.2` directory and use it accordingly.

[...]

PR #142, opened 2023-02-17
Add doc about classpathResources + multiple versions

mike-solomon:
Fixes: https://github.com/openrewrite/rewrite-docs/issues/124

### DIFF
--- a/authoring-recipes/multiple-versions.md
+++ b/authoring-recipes/multiple-versions.md
@@ -43,8 +43,8 @@ With the `recipe-library` plugin added, you can then add a `recipeDependencies` 
 recipeDependencies {
     parserClasspath("org.mockito:mockito-core:2.16.0")
     parserClasspath("org.mockito:mockito-core:3.12.4")
-    parserClasspath("unit:junit:3.7")
-    parserClasspath("unit:junit:4.13.2")
+    parserClasspath("junit:junit:3.7")
+    parserClasspath("junit:junit:4.13.2")
 }
 ```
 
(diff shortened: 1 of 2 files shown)


authoring-recipes/multiple-versions.md
```diff
@@ -0,0 +1,50 @@
...
+}
+```
+
+With the `recipe-library` plugin added, you can then add a `recipeDependencies` section in your `build.gradle` file. In there, you can add a `parserClasspath` that specifies a dependency and its corresponding version:
+
+```groovy
+recipeDependencies {
+    parserClasspath("org.mockito:mockito-all:1.10.19")
+    parserClasspath("org.mockito:mockito-core:3.+")
+    parserClasspath("org.mockito:mockito-junit-jupiter:3.+")
```
timtebeek, line 46: I'd maybe keep this in line with the `.classpathFromResources("junit-4.13.2", "mockito-core-3.12.4"));` above, and maybe use two different versions of the same artifact, with the same pinned versions. That makes it a bit more clear this is intended for different versions of the same artifact that would otherwise conflict. Not using version ranges here might help prevent multiple versions from the same range ending up in a repository.

But open to alternative views on all of this!

mike-solomon: Good suggestion. 

One question, though: What do you mean by "two different versions of the same artifact, with the same pinned versions?"

mike-solomon: Also - does the new commit help resolve your concerns or did I miss something on that front?

timtebeek (approved):

timtebeek:
Nicely done!

PR #143, opened 2023-02-19
Document method pattern varargs examples

timtebeek:
This came up while working on a recipe that tried to match a specific invocation with two arguments, of a method that takes a variable number of arguments in https://github.com/openrewrite/rewrite/issues/2859#issuecomment-1435650842 . Figured document the case in the table so there's no confusion about matching method invocations.

### DIFF
--- a/reference/method-patterns.md
+++ b/reference/method-patterns.md
@@ -29,21 +29,27 @@ The return type of a method is _not_ represented in method patterns. Methods in 
 
 ### Examples
 
-So if you wanted to match invocations of `String.toString(int beginIndex)` then this expression would match only that method: `java.lang.String toString(int)`. If you wanted to match the two-argument form of the same
+So if you wanted to match invocations of `String.substring(int beginIndex)` then this expression would match only that method: `java.lang.String substring(int)`.
+If you wanted to match the two-argument form of the same, you can use the method pattern `java.lang.String substring(int, int)`.
+The below table shows some more examples of method patterns and the methods they match.
 
-| Method Pattern                              | Matches                                                                                                                                                                                              |
-|:--------------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `java.lang.String substring(int)`           | Exactly the [single argument overload](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#substring%28int%29) of `String.substring()`                                |
-| `java.lang.String substring(int, int)`      | Exactly the [two argument overload](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#substring%28int,int%29) of `String.substring()`                               |
-| `java.lang.String substring(..)`            | Any overload of `String.substring()`                                                                                                                                                                 |
-| `java.lang.String *(int)`                   | Any method on `String` that accepts a single argument of type `int`                                                                                                                                  |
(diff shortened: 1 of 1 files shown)


reference/method-patterns.md
```diff
@@ -29,19 +29,27 @@ The return type of a method is _not_ represented in method patterns. Methods in
...
-| `java.lang.String substring(..)` | Any overload of `String.substring()` |
-| `java.lang.String *(int)` | Any method on `String` that accepts a single argument of type `int` |
-| `com.yourorg.Foo bar(int, String, ..)` | Any method on `Foo` named `bar` accepting an `int`, a `String`, and zero or more other arguments of any type |
-| `com.yourorg.Foo <constructor>(*, *, *)` | Constructors of `Foo` accepting exactly three arguments of any type. |
-| `*..String *(..)` | Any method accepting any arguments on classes named "String" in any package |
-| `*..* *(..)` | Any method accepting any arguments on any class |
-| `org.example..* *(..)` | Any method on any class in the "org.example" package, or any sub-package of "org.example" |
+So if you wanted to match invocations of `String.substring(int beginIndex)` then this expression would match only that method: `java.lang.String substring(int)`.
+If you wanted to match the two-argument form of the same, you can use the method pattern `java.lang.String substring(int, int)`.
```
timtebeek, line 33: There seems to have been a half-finished sentence there; Wrapped it up as best as I could, and swapped out `toString` for `substring`, as `toString` does not take any `int` arguments.

My suggestion though would be to just drop these two sentences and instead introduce the table, if we even do that at all.
```suggestion
```

reference/method-patterns.md
```diff
@@ -29,19 +29,27 @@ The return type of a method is _not_ represented in method patterns. Methods in
...
+| `*..String *(..)`                           | Any method accepting any arguments on classes named "String" in any package                                                                                                                                     |
+| `*..* *(..)`                                | Any method accepting any arguments on any class                                                                                                                                                                 |
+| `org.example..* *(..)`                      | Any method on any class in the "org.example" package, or any sub-package of "org.example"                                                                                                                       |
+
+{% hint style="info" %}
+When matching methods using varags, method patterns should match the parameter _declaration_, not the parameter _use_ for method invocations.
```
timtebeek, line 51: Figured an extra comment on matching varargs couldn't hurt. Didn't fit well into the table above.

timtebeek: Might even expand this with something like:

A method can be declared as `org.apache.commons.lang3.Validate notNull(Object, String, Object[])`, but can be invoked with a variable number of arguments.
```
Validate.notNull(obj, "Obj was null")
Validate.notNull(obj, "Obj %s was null", "foo")
Validate.notNull(obj, "Obj %s null in %", "foo", "bar")
```
When matching any of the above method invocations, use the method declaration rather than the actual number of arguments.

mike-solomon (approved):
Changes look good to me. Thanks for doing this :)

reference/method-patterns.md
mike-solomon: Making the table makes sense to me.

timtebeek: Oh the table was already there, I merely expanded it, which kind of butchered the diff.

I'll remove the introduction sentences, which were half finished to begin with, and duplicate the table. :)

reference/method-patterns.md
sambsnyd: @timtebeek this hint is a bit misleading. `MethodMatcher` _always_ match against declared parameters, not supplied parameters. You're matching the method _being_ invoked, not the method invocation itself, if that makes sense.

PR #153, opened 2023-03-20
Also update rewrite-maven-plugin in sync and document pluginRepositories

timtebeek:
Tried this locally, and found that in some cases we also need to use snapshot versions of the Rewrite Maven plugin. Figured document that as well, as it also needs `<pluginRepositories>`, and then just go ahead and recommend people keep those in sync in case there's breaking changes in openrewrite/rewrite that are needed in the module snapshots.


mike-solomon (approved):
LGTM

PR #158, opened 2023-03-29
Change the left panel of the root page to be expandable

kunli2:
Currently, on https://docs.openrewrite.org/ the left panel is a long list and not expandable.

This PR changes the left panel of the root page to be expandable and the view should be more clear and well orgnaized.

Notes:
I tried to use run `gitbook` server locally and verify, however, I can not run `gitbook serve` command and it's stuck somehow. so can not provide the view after the change.


mike-solomon:
Hard for me to know without seeing this in action. I’m not sure if these changes would do what is expected. +1 to Sam’s suggestion for this. I think visual structure changes like this would be best to try and do in the GitBook UI itself so we can see. 

If you go to our Gitbook page, you can just make edits right in the browser and then share the link with us.

Another option would be to test in our Knowledge base docs by doing this on a smaller scale since those docs are private.

kunli2:
Thanks for the inputs, I tried Gitbook editor interface, however, I think it supports editing the content but seems don't support changing the formatting of the left panel...

kunli2:
Sorry, I falsely merged a test PR on GitBook UI (which has been reverted), and the commit was  https://github.com/openrewrite/rewrite-docs/commit/e2eb20e721fbf534e00f4b8337663238844bc3db 

However, we can see when adding a page and sub-pages like this, 
1. it's folded by default
Screen Shot 2023-03-29 at 5 07 10 PM
2. expandable to
Screen Shot 2023-03-29 at 5 07 30 PM


the according code change in git log is like
```md
## Running Recipes

* [Page 2](running-recipes/page-2/README.md)
  * [Sub-Page 2.1](running-recipes/page-2/sub-page-2.1.md)
* [Page 1](running-recipes/page-1/README.md)
  * [Sub-Page 1.1](running-recipes/page-1/sub-page-1.1.md)
  * [Sub-Page 1.2](running-recipes/page-1/sub-page-1.2.md)
```

So I think the format in this PR is correct, and we can imagine this change will make the left panel folded by category by default, and it's expandable.

mike-solomon (approved):
I’m down with merging this in and seeing how it looks and whether people like it or not. Can always adjust based on feedback. Thanks for thinking about the docs @kunli2 ^_^

PR #171, opened 2023-05-24
Bump rewrite-maven-plugin and rewrite-spring version

timtebeek:
In response to https://stackoverflow.com/questions/76322306/how-can-i-fix-the-nosuchmethoderror-while-using-the-open-rewrite-plugin-for-migr

### DIFF
--- a/authoring-recipes/recipe-development-environment.md
+++ b/authoring-recipes/recipe-development-environment.md
@@ -323,7 +323,7 @@ Now you can run your recipe with `./gradlew rewriteRun` or `./gradlew rewriteDry
             <plugin>
                 <groupId>org.openrewrite.maven</groupId>
                 <artifactId>rewrite-maven-plugin</artifactId>
-                <version>4.44.0</version>
+                <version>4.46.0</version>
                 <configuration>
                     <activeRecipes>
                         <recipe> [your recipe name] </recipe>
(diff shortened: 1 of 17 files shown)


timtebeek:
@mike-solomon I know you usually go through these periodically to upgrade; I figured do a quick fix now as users are running into these issues following a recent breaking change.

Alternatively we can swap out these explicit version numbers with `RELEASE`, such that it should always work. We already somewhat use that in our reference recipes for the Maven commanline tab:
https://docs.openrewrite.org/recipes/java/migrate/upgradetojava17

Note the absence of version numbers here:
```bash
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE \
  -Drewrite.activeRecipes=org.openrewrite.java.migrate.UpgradeToJava17
```

But not yet for the Maven tab:
```xml
<project>
  <build>
    <plugins>
      <plugin>
        <groupId>org.openrewrite.maven</groupId>
        <artifactId>rewrite-maven-plugin</artifactId>
        <version>4.45.0</version>
        <configuration>
          <activeRecipes>
            <recipe>org.openrewrite.java.migrate.UpgradeToJava17</recipe>
          </activeRecipes>
        </configuration>
        <dependencies>
          <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>1.21.1</version>
          </dependency>
        </dependencies>
      </plugin>
    </plugins>
  </build>
</project>
```

By standardizing on version-less docs where we can, our users are less likely to use older versions, and we don't have to update as many parts of the docs as often. The downside is that it's [less explicit about version numbers](https://github.com/openrewrite/rewrite-gradle-plugin/issues/199).

What are your thoughts?

timtebeek:
> Is this generated? Why is it out-of-date in that case?

Only https://github.com/openrewrite/rewrite-docs/tree/master/reference/recipes is generated; these changes are outside that directory, as the generated docs already use 4.45.0 for maven: https://github.com/openrewrite/rewrite-docs/blob/master/reference/recipes/java/migrate/javaversion17.md

Mike has a process for also updating the generated docs, which I intentionally left out here.
I want the recommended guides to mostly work, which isn't the case right now unfortunately.

PR #172, opened 2023-06-07
Add check-spelling

jsoref:
## What's changed?

This introduces a check-spelling workflow, roughly as I used it to develop the PR, but with a couple of changes.

As configured check-spelling will produce a job summary reachable via [Details](https://github.com/jsoref/rewrite-docs/actions/runs/4993724999#summary-13522925151).

When possible, it will generate a Sarif report, and if a user is an admin (or nearly an admin) they'll be able to view it in GitHub, otherwise, users can download a generated artifact and process w/ their own tools (especially VSCode).

## What's your motivation?

Typos make it hard to read / parse content as well as search for content. Catching them before they're merged makes my life easier as a consumer of repositories.

@timtebeek @mike-solomon 

I can run this periodically in my own forks -- but I'm unlikely to revisit a given repository more frequently than annually if I do (more likely in 3-5 years).

## Any additional context

I've configured it to provide a fallback message (using `spell_check_this: openrewrite/rewrite-docs@master`) (ideally this shouldn't be necessary, but I think I have a bug I need to resolve which I haven't had time to focus on).

The `advice.md` file is a markdown file that the project can customize based on the needs of its maintainers/contributors. The `README.md` is designed to be helpful documentation, but is absolutely optional.

You'll want to review `expect.txt` at least once, and potentially periodically (depending on how good people are at paying attention to the tool -- at work people sometimes just accept its suggestions even when it's pointing out obvious typos 🤷‍♂️). 

The update job is designed to only run in forks. It enables people to talk to their own instance instead of running the update command locally / manually editing `expect.txt` / `excludes.txt`. You're encouraged to play with this logic in a fork, or if you aren't comfortable with it, there's nothing wrong with omitting the job.

It's using the latest released tag ([v0.0.21](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.21)). Some projects are release-early / release-often -- check-spelling isn't. I don't tend to make a lot of releases although my coworkers are dogfooding [prerelease](https://github.com/check-spelling/check-spelling/tree/prerelease) in a dozen repositories (and I'm constantly dogfooding it to make contributions such as the one that was merged here). As you're already using GitHub workflows, you'll want to set up dependabot to make PRs for updates to things.

I'm happy to answer questions here / in Slack, and if this is merged, I'll probably watch the repository for a bit to make sure people don't have any problems.

### Checklist
- [ ] I've used the IntelliJ auto-formatter on affected files
- [ ] I've updated the documentation (if applicable)


SUMMARY.md
```diff
@@ -811,7 +811,7 @@
         * [Use MatcherAssert#assertThat(..)](reference/recipes/java/testing/junit5/usehamcrestassertthat.md)
         * [Use Mockito JUnit Jupiter extension](reference/recipes/java/testing/junit5/usemockitoextension.md)
         * [Use OkHttp 3 MockWebServer for JUnit 5](reference/recipes/java/testing/junit5/upgradeokhttpmockwebserver.md)
-        * [Use Vertx JUnit 5 Extension](reference/recipes/java/testing/junit5/vertxunittovertxjunit5.md)
+        * [Use Vertex JUnit 5 Extension](reference/recipes/java/testing/junit5/vertxunittovertxjunit5.md)
```
timtebeek, line 814: This refers to https://github.com/eclipse-vertx/vertx-junit5/blob/master/src/main/java/io/vertx/junit5/VertxExtension.java
```suggestion
        * [Use Vert.x JUnit 5 Extension](reference/recipes/java/testing/junit5/vertxunittovertxjunit5.md)
```

SUMMARY.md
```diff
@@ -811,7 +811,7 @@
         * [Use MatcherAssert#assertThat(..)](reference/recipes/java/testing/junit5/usehamcrestassertthat.md)
         * [Use Mockito JUnit Jupiter extension](reference/recipes/java/testing/junit5/usemockitoextension.md)
         * [Use OkHttp 3 MockWebServer for JUnit 5](reference/recipes/java/testing/junit5/upgradeokhttpmockwebserver.md)
-        * [Use Vertx JUnit 5 Extension](reference/recipes/java/testing/junit5/vertxunittovertxjunit5.md)
+        * [Use Vert.x JUnit 5 Extension](reference/recipes/java/testing/junit5/vertxunittovertxjunit5.md)
```
mike-solomon, line 814: Not a big deal - but as these are referencing the `reference/recipes` files, this will just get clobbered every release (as this part of the `SUMMARY.md` is overwritten). Would be helpful to leave check spelling on for the rest of the file, though.

mike-solomon (approved):
This looks great. Thanks so much for taking the time to do this.

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,145 @@
...
+    concurrency:
+      group: spelling-update-${{ github.event.issue.number }}
+      cancel-in-progress: false
+    steps:
+    - name: apply spelling updates
+      uses: check-spelling/check-spelling@v0.0.21
+      with:
+        experimental_apply_changes_via_bot: 1
+        checkout: true
+        ssh_key: "${{ secrets.CHECK_SPELLING }}"
```
timtebeek, line 145: So if I understand correctly this part is only if we wish to update some of these files through a fork, right?

SUMMARY.md
timtebeek: We can [fix this at the source](https://github.com/openrewrite/rewrite-testing-frameworks/blob/ee9a439948dea63fb3d117f1f0bac39d80e50670/src/main/resources/META-INF/rewrite/junit5.yml#L211-L212) indeed; good that you noticed!

.github/workflows/spelling.yml
jsoref: Right.

To trigger it:
1. Create your own personal fork
2. In the fork, create a branch with a typo (or an unknown word...)
3. Create a pull request into the default branch of the fork.
4. Go to the ❌ for the status and follow the instructions. The first round through, it'll walk you through creating a deploy key (which is optional) -- if you do that, the second time you use the feature, it will be a bit smoother.

Issue #141, opened 2023-02-17
Document supported languages on docs.openrewrite.org


timtebeek:
We had a question come in on Slack:

> Hi, What other languages does OpenRewrite work with?

Which intuitively at first seemed strange; as "we all know" what languages we support.
Yet it's not listed explicitly in the documentation anywhere, and could be helpful to add.

The answer given was:
```
We have support for:
Programming languages:
-  Java
-  Groovy/Gradle - Focus is on supporting Gradle dependency manipulation, but works on most regular groovy sources as well
-  Kotlin - Early support has been released, but it is quite new, still being actively developed and not widely tested
-  Python - Not yet released. In development
-  COBOL - Closed source, commercial feature
Data formats:
-  XML
-  Properties
-  YAML
-  JSON
-  Protobuf
-  Maven POMs, with dependency maanagement
After finishing Kotlin and Python we will next expand into Typescript and C#
```

Putting that on a page somewhere, or even a landing page, could help those that quickly want to evaluate OpenRewrite.

mlvandijk:
For Python, it would be interesting to know how to run those recipes. Currently the documentation is Maven/Gradle (JVM) focussed

mlvandijk:
Also Kotlin is not really (well) represented in the documentation (yet).

timtebeek:
To close this out: the page should start to show on: https://docs.openrewrite.org/reference/supported-languages

PR #174, opened 2023-06-08
Adds a doc for all supported languages

mike-solomon:
This is a basic start for this doc. @timtebeek if there's anything else you think I should call out here please let me know.

Fixes: https://github.com/openrewrite/rewrite-docs/issues/141

### DIFF
--- a/introduction.md
+++ b/introduction.md
@@ -4,7 +4,7 @@
 
 ## **Large-scale automated source code refactoring**
 
-OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language ([see supported languages](/reference/supported-languages.md)).
+OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations. While the original focus was on the Java language, OpenRewrite is [continuously expanding language and framework coverage](/reference/supported-languages.md).
 
 ## Semantic code search and transformation
 
--- a/introduction.md
+++ b/introduction.md
@@ -4,7 +4,7 @@
 
 ## **Large-scale automated source code refactoring**
 
-OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language ([all supported languages](/reference/supported-languages.md)).
+OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language ([see supported languages](/reference/supported-languages.md)).
 
 ## Semantic code search and transformation
 
--- a/introduction.md
+++ b/introduction.md
@@ -4,7 +4,7 @@
 
 ## **Large-scale automated source code refactoring**
 
-OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language.
+OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language ([all supported languages](/reference/supported-languages.md)).
 
 ## Semantic code search and transformation
 
--- a/reference/supported-languages.md
+++ b/reference/supported-languages.md
@@ -28,4 +28,4 @@ Framework migration recipes are developed and maintained in collaboration of the
 * Jakarta
 * Kubernetes
 
-You can find all of the frameworks in the [community marketplace](https://public.moderne.io/marketplace).
+You can find all of the frameworks in the [recipe catalog](https://docs.openrewrite.org/recipes).
--- a/reference/supported-languages.md
+++ b/reference/supported-languages.md
@@ -1,0 +1,30 @@
+# Languages/Formats OpenRewrite supports
+
+OpenRewrite is continuously adding support for new languages and formats. As of June 2023, OpenRewrite supports:
+
+## Programming languages
+
+* Java
+* Groovy/Gradle 
+* Kotlin
+* COBOL (Closed source, commercial feature)
+
+## Data formats
+
+* XML
+* Properties
+* YAML
+* JSON
+* Protobuf
+* Maven POMs (with dependency management)
+
+## Frameworks
+
+Framework developers have created recipes to help migrate and support their frameworks. These include:
+
+* Spring
+* Quarkus
+* Micronaut
+* Kubernetes
+
+You can find all of the frameworks in the [community marketplace](https://public.moderne.io/marketplace).
--- a/reference/supported-languages.md
+++ b/reference/supported-languages.md
@@ -20,7 +20,7 @@ OpenRewrite is continuously adding support for new languages and formats. As of 
 
 ## Frameworks
 
-Framework developers have created recipes to help migrate and support their frameworks. These include:
+Framework migration recipes are developed and maintained in collaboration of the team behind OpenRewrite at Moderne, the original framework authors, and the wider OSS community. Individual recipes list the contributing authors. Support includes, but is not limited to:
 
 * Spring
 * Quarkus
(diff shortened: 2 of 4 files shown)


reference/supported-languages.md
```diff
@@ -0,0 +1,30 @@
...
+* XML
+* Properties
+* YAML
+* JSON
+* Protobuf
+* Maven POMs (with dependency management)
+
+## Frameworks
+
+Framework developers have created recipes to help migrate and support their frameworks. These include:
```
mike-solomon, line 23: No idea if this is the right way of phrasing this. Would appreciate any advice on how this might be phrased differently.

jsoref: Is this sentence necessary? There isn't an equivalent one for Programming languages nor for Data formats.

mike-solomon: I was asked to call out that framework authors are responsible for these frameworks and made these rather than OpenRewrite/Moderne.

jsoref, line 23:
```suggestion
Framework developers can create recipes to help migrate and support their frameworks. Supported frameworks include:
```

reference/supported-languages.md
```diff
@@ -0,0 +1,30 @@
+# Languages/Formats OpenRewrite supports
+
+OpenRewrite is continuously adding support for new languages and formats. As of June 2023, OpenRewrite supports:
+
+## Programming languages
+
+* Java
+* Groovy/Gradle 
+* Kotlin
+* COBOL (Closed source, commercial feature)
```
timtebeek, line 10:
```suggestion
* COBOL (open source, but not free for commercial use)
```

introduction.md
```diff
@@ -4,7 +4,7 @@
 
 ## **Large-scale automated source code refactoring**
 
-OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language.
+OpenRewrite enables large-scale distributed source code refactoring for framework migrations, vulnerability patches, and API migrations with an early focus on the Java language ([see supported languages](/reference/supported-languages.md)).
```
timtebeek, line 7: When I scan-read this sentence, even with the `see supported languages`, I might still walk away thinking there's no support for anything other than Java, which I don't think reflects our recent efforts. Can we update this to for instance say somehting along the lines of

> early focus on the Java language, and [continuously expanding language coverage](/reference/supported-languages.md)

reference/supported-languages.md
timtebeek: Based off on https://github.com/openrewrite/rewrite-cobol , which uses this wording as well. If changed here then perhaps we'd best be consistent there as well.

reference/supported-languages.md
timtebeek: I'd go with something like:
```suggestion
Framework migration recipes are developed and maintained in collaboration of the team behind OpenRewrite at Moderne, the original framework authors, and the wider OSS community. Individual recipes list the contributing authors. Support includes, but is not limited to:
```

This softly calls to action anyone looking to contribute, such that it's clear we're open to contributions. We also credit anyone that contributed to any recipe, such that there's no confusion who's involved.

reference/supported-languages.md
```diff
@@ -0,0 +1,30 @@
...
+* Maven POMs (with dependency management)
+
+## Frameworks
+
+Framework developers have created recipes to help migrate and support their frameworks. These include:
+
+* Spring
+* Quarkus
+* Micronaut
+* Kubernetes
```
timtebeek, line 28: I'd include Jakarta here as well, as that's a common migration users might be looking for these days.
```suggestion
* Jakarta
* Kubernetes
```
I'm somewhat hesitant about including Kubernetes; we haven't explicitly documented how to run such migration recipes, nor do we hear much from users that are using these recipes.

reference/supported-languages.md
```diff
@@ -0,0 +1,30 @@
...
+
+You can find all of the frameworks in the [community marketplace](https://public.moderne.io/marketplace).
```
timtebeek, line 30: Since these are the OSS docs, should we not link to the recipe catalog instead of the marketplace? Both contain the same, and while the marketplace looks better, it might still be slightly confusing for readers of the OSS docs to be transported elsewhere.
```suggestion
You can find all of the frameworks in the [recipe catalog](https://docs.openrewrite.org/recipes).
```

introduction.md
mike-solomon: Yep will do - thanks!

reference/supported-languages.md
mike-solomon: Olga had mentioned that she wanted to call out the marketplace -- but I think there may just be some context missing. Agree that it's probably better to just link to the recipe catalog.

mike-solomon:
@timtebeek rephrased a few pieces - would appreciate you taking another look when you have a moment :D

reference/supported-languages.md
```diff
@@ -0,0 +1,36 @@
+# Languages/Formats OpenRewrite supports
+
+OpenRewrite is continuously adding support for new languages and formats. As of June 2023, OpenRewrite supports:
+
+## Programming languages
+
+* Java
+* Groovy/Gradle 
+* Kotlin
```
timtebeek, line 9:
```suggestion
* Kotlin
* Groovy
```
Gradle isn't as much a language as it is a build tool, much like Maven. It might be slightly confusing to have Gradle mentioned here.
And Kotlin is gaining traction, so figured list that above Groovy to attract the youngsters. ;)

mike-solomon: haha fair enough :D

reference/supported-languages.md
```diff
@@ -0,0 +1,36 @@
...
+* COBOL (open-source, but not free for commercial use)
+
+## Data formats
+
+* XML
+* Properties
+* YAML
+* JSON
+* Protobuf
+* Maven POMs (with dependency management)
```
timtebeek, line 19: We could pull Maven and Gradle into a separate block, with Bazel soon to be added, and possibly tools like NPM as well.
```suggestion

## Build tools
* Maven
* Gradle
```

reference/supported-languages.md
```diff
@@ -0,0 +1,36 @@
+# Languages/Formats OpenRewrite supports
+
+OpenRewrite is continuously adding support for new languages and formats. As of June 2023, OpenRewrite supports:
+
+## Programming languages
+
+* Java
+* Kotlin
+* Groovy
+* COBOL (open-source, but not free for commercial use)
```
timtebeek, line 10: Is it too early to add Python & TS/JS? They're pretty close...
```suggestion
* COBOL (open-source, but not free for commercial use)
* Python
* TypeScript / JavaScript
```

mike-solomon: Should we include Python and TypeScript even though those are only sort of supported right now?

timtebeek (approved):
Approved already; good to have something like this going already..

If we want to be fancy we can lift over some of the graphics that we have for marketing
image

reference/supported-languages.md
mike-solomon: I'm unsure -- I left them out because it felt like we didn't _really_ support them, yet. 

Tracey said: 

```
Python is just a simple text search for now; it’s not functional yet.
TypeScript is in the early stages of support.
```

mike-solomon:
> Approved already; good to have something like this going already..
> 
> If we want to be fancy we can lift over some of the graphics that we have for marketing image

Fair suggestion - do you know where that graphic is from? I think it's sort of funny that Micronaut has a misspelling line under the name. Would want to fix that before we add it.

mike-solomon:
Gonna merge this in as is for now - but open to adding more after. Want to check in with Olga on whether or not we should include Python and TypeScript.

PR #175, opened 2023-06-09
Ignore drafts for check-spelling

jsoref:
## What's changed?

Check Spelling will not perform checks for pull-requests while they're in draft state.

There are two sides to this:
1. An additional `if:` condition for `github.event.pull_request.draft == false` -- this suppresses checks while the PR is in draft (note that any initial `push` events will still run and may trigger scanning / reporting if it's unhappy, but subsequent runs once a PR is opened will be suppressed by check-spelling's `suppress_push_for_open_pull_request`)
2. `on:` / `pull_request_target:` / `types:` / `'ready_for_review'` -- this triggers a check when someone converts the PR from draft to ready for review (otherwise you'd have to wait for an additional push, which would be frustrating).

## What's your motivation?

@mike-solomon asked to avoid being told about spelling while his PR (#167) was in draft

### Checklist
- [ ] I've used the IntelliJ auto-formatter on affected files
- [ ] I've updated the documentation (if applicable)

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -69,6 +69,7 @@ on:
     - 'opened'
     - 'reopened'
     - 'synchronize'
+    - 'ready_for_review'
   issue_comment:
     types:
     - 'created'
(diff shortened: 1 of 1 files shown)


mike-solomon (approved):
You rock -- thanks for making this change.

kunli2 (approved):

PR #183, opened 2023-06-21
Cleanup visitors doc for rewrite8

mike-solomon:
I didn't really see many changes that needed to be made here and I'm unsure if that's because none of this stuff changed or if we are missing some key pieces.

I did find a few minor cleanup pieces and I did call out that one of the examples doesn't work. I think I'll need some help on that front.

### DIFF
--- a/concepts-and-explanations/visitors.md
+++ b/concepts-and-explanations/visitors.md
@@ -45,7 +45,7 @@ As an example of how the `Cursor` can be helpful, imagine a visitor that is task
 
 ```java
 @Override
-public J visitClassDeclaration(J.ClassDeclaration cd, ExecutionContext context) {
+public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration cd, ExecutionContext context) {
     // The base class provides the language-specific navigation of sub-elements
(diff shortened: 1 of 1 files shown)


concepts-and-explanations/visitors.md
```diff
@@ -164,6 +164,8 @@ There are some cases, though, where you may want to share other types of informa
 
 In those cases, you can specify a different, mutable type for your visitor such as in this visitor which counts the number of method invocations in a Java LST:
 
+TODO: This does not work for me. I get errors about `static` declarations not being allowed in inner classes and also that the visitor can't be referenced from a static context.
```
jsoref, line 167:
```suggestion
TODO(mike-solomon): This does not work for me. I get errors about `static` declarations not being allowed in inner classes and also that the visitor can't be referenced from a static context.
```

mike-solomon: This isn't going to be committed into the final doc. Was adding this to get feedback from reviewers. Thanks for the suggestion, though!

kunli2: LGTM, Yeah, we can remove this todo and merge,

kunli2 (approved):

Issue #190, opened 2023-06-29
Checkstyle doc has incorrect versions 


mike-solomon:
[As reported by a community member](https://rewriteoss.slack.com/archives/C01A843MWG5/p1688021030190299), our current checkstyle doc doesn't appear to use the correct versions and, therefore, doesn't work. 

It appears that not only has the recipe been moved, but the versions are out of date. We need to ensure that if someone follows that documentation that it works as expected.

PR #193, opened 2023-06-29
Fix the checkstyle recipe guide

mike-solomon:
Fixes: https://github.com/openrewrite/rewrite-docs/issues/190

Right now, the checkstyle guide doesn't work. I _think_ I've fixed all the issues here but would appreciate a second pair of eyes on the code.


sambsnyd (approved):

timtebeek (approved):
Looks plausible to me; haven't ran it, but think this ought to work.

PR #194, opened 2023-06-29
Simplify the Gradle initscript in use for running without modification to the project`

shanman190:


marcorichetta:
Is this outdated? https://docs.openrewrite.org/running-recipes/running-rewrite-on-a-gradle-project-without-modifying-the-build#step-2-create-a-gradle-init-script.

I tried to execute it **as is** in a gradle project and I got the error:

```shell
Execution failed for task ':rewriteRun'.
> org.openrewrite.RecipeException: Recipes not found: org.openrewrite.FindSpringUses
```

- There still is a reference to `allprojects` which was removed from `init.gradle`
https://github.com/openrewrite/rewrite-docs/blob/378254f389d8c8c2ca654543fb4de10212998f16/running-recipes/running-rewrite-on-a-gradle-project-without-modifying-the-build.md?plain=1#L13

- The reference to `rewrite.yaml` file was removed from the script, so how is it used?
https://github.com/openrewrite/rewrite-docs/pull/194/files#diff-d48e469124b4afe3ef6d2521dfd18ce136fa748096931c02f31402dc3277238bL38

shanman190:
So you are correct that the text is now out of sync with the code snippet itself. I'll get a fix out for that momentarily.

The `rewrite.yml` is the default file name, so the property was really just setting the same default again, so that's why it was removed so as to make the getting started process a little simpler, but that should probably be stated a little more explicitly.

shanman190:
@marcorichetta, docs have been fixed up in the repository. They'll be deployed to GitBook shortly.

PR #195, opened 2023-07-04
update pathExpressions in example to list

pstreef:
## What's changed?
Documentation

## What's your motivation?
The example does not work


timtebeek:
Thanks a lot for the fix!

Issue #191, opened 2023-06-29
Add a doc for upcoming language support


mike-solomon:
Various people have requested that we share what languages we plan on supporting in the near future in OpenRewrite. We should consider making a public page that contains this information. 

[Related Slack thread](https://moderneinc.slack.com/archives/C01VADFPJQZ/p1687972532080459)

mike-solomon:
The open PR for this is blocked right now as Olga and the team wanted more time to ponder this doc.

mike-solomon:
Closing this as it was requested I move this to the Moderne docs instead. You can find this [here](https://docs.moderne.io/references/supported-languages)

PR #198, opened 2023-07-10
Add future plans to supported languages

mike-solomon:
Fixes: https://github.com/openrewrite/rewrite-docs/issues/191


mike-solomon:
It was requested that I remove this doc and put it in the Moderne docs instead. This is partially due to the fact that a lot of these languages are really only supported in Moderne and not in OpenRewrite like I originally thought. Closing this PR

Issue #129, opened 2023-02-02
Document how to work with Maven multi module projects


timtebeek:
We've had some instances [in the past](https://github.com/openrewrite/rewrite-maven-plugin/issues/103) where [people asked](https://github.com/openrewrite/rewrite-maven-plugin/issues/475#issuecomment-1387620507) about the [best way](https://rewriteoss.slack.com/archives/C01AB6L98TC/p1674032746612539) to [run for a Maven multi module project](https://stackoverflow.com/questions/75148939/rewrite-java-packages-and-groupid-artifact-names-of-super-pom-project).

While ideally we improve the way things currently work, for the time being it could be good to document a way around the issue through a [Maven profile in the parent pom.xml](https://github.com/openrewrite/rewrite-maven-plugin/issues/475#issuecomment-1387620507). That way people can find (or be directed to) a way around the problem.

PR #201, opened 2023-07-11
Add short explanation of multi-module maven + rewrite

mike-solomon:
Fixes: https://github.com/openrewrite/rewrite-docs/issues/129

I'm not actually sure if there's anything more I should add here or if I even described the problem very well. 

@timtebeek Would greatly appreciate if you could take a look and let me know if there's anything else worth adding here.

### DIFF
--- a/running-recipes/multi-module-maven.md
+++ b/running-recipes/multi-module-maven.md
@@ -26,3 +26,5 @@ In the meantime, one way to work around this is to add a profile on the parent P
   </build>
 </profile>
 ```
+
+You can then run OpenRewrite from any submodule using this command: `mvn -Popenrewrite rewrite:run`. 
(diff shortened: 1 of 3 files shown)


running-recipes/multi-module-maven.md
```diff
@@ -0,0 +1,28 @@
...
+          <activeRecipes>
+            <recipe>org.openrewrite.java.cleanup.Cleanup</recipe>
+          </activeRecipes>
+          <failOnDryRunResults>true</failOnDryRunResults>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+</profile>
+```
```
timtebeek, line 28: I'd add here how people can activate the profile and run the recipes for those less familiar with Maven profiles. Something like

> You can then run OpenRewrite from any submodule using `mvn -Popenrewrite rewrite:run`.

Haven't tested that, but I think it ought to work.

mike-solomon: Updated. Gonna merge this in for now as it's past your end of day - if my changes are wrong or you think I should make additional updates please let me know. Happy to make more changes.

PR #216, opened 2023-07-31
doc: document why the BOM is not mentioned

SimonVerhoeven:
## What's changed?

I've documented why the BOM isn't mentioned

## What's your motivation?

To improve the documentation, and since people were pondering the why on Slack.

@mike-solomon 

no

### Checklist
- [x] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/running-recipes/getting-started.md
+++ b/running-recipes/getting-started.md
@@ -298,7 +298,7 @@ After doing that, your `pom.xml` file should look similar to this:
 
 To double-check that everything is working, run the command `./mvnw rewrite:run`. Your project should be upgraded to Spring Boot 2 and all of the test classes should be updated to JUnit 5. Your `pom.xml` file will also have had its Spring dependencies updated, the JUnit 4 dependency removed, and the JUnit 5 dependency added.
 
-__note__ here [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom\_import) is not mentioned for `dependencyManagement` as we can not use it to version control the OpenRewrite plugin dependencies, unlike with the Gradle setup below.
+__note__ here [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom\_import) is not mentioned for `dependencyManagement` as we cannot use it to version control the OpenRewrite plugin dependencies, unlike with the Gradle setup below.
 
 ### Gradle + external modules
 
--- a/running-recipes/getting-started.md
+++ b/running-recipes/getting-started.md
@@ -298,6 +298,8 @@ After doing that, your `pom.xml` file should look similar to this:
 
 To double-check that everything is working, run the command `./mvnw rewrite:run`. Your project should be upgraded to Spring Boot 2 and all of the test classes should be updated to JUnit 5. Your `pom.xml` file will also have had its Spring dependencies updated, the JUnit 4 dependency removed, and the JUnit 5 dependency added.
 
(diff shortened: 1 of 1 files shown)


running-recipes/getting-started.md
```diff
@@ -298,6 +298,8 @@ After doing that, your `pom.xml` file should look similar to this:
 
 To double-check that everything is working, run the command `./mvnw rewrite:run`. Your project should be upgraded to Spring Boot 2 and all of the test classes should be updated to JUnit 5. Your `pom.xml` file will also have had its Spring dependencies updated, the JUnit 4 dependency removed, and the JUnit 5 dependency added.
 
+__note__ here [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom\_import) is not mentioned for `dependencyManagement` as we can not use it to version control the OpenRewrite plugin dependencies, unlike with the Gradle setup below.
```
jsoref, line 301:
```suggestion
__note__ here [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom\_import) is not mentioned for `dependencyManagement` as we cannot use it to version control the OpenRewrite plugin dependencies, unlike with the Gradle setup below.
```

I'm curious about `here` (It's often an [a11y faux pas](https://usability.yale.edu/web-accessibility/articles/links#:~:text=Avoid%20link%20text%20like%20%E2%80%9CClick,experience%20with%20duplicated%20link%20text.))

SimonVerhoeven: But `here` isn't the link text, `bill of materials (BOM) dependency` is.

mike-solomon: Thanks for taking the time to write this!

Does this rephrase make sense to you? I wasn't sure if there was a benefit to linking to the Gradle BOM here or if calling out `dependencyManagement` would be useful. However, I could definitely be missing something. Open to editing this further!

Also, separately, GitBook provides a code snippet we can use to get stylized hints in the docs. I added that to be suggestion below.

```suggestion
{% hint style="info" %}
Maven does not currently support using a bill of materials (BOM) to specify plugin versions or dependencies. This means that you will have to specify the versions of each plugin by hand, unlike in the Gradle section below.
{% endhint %}
```

mike-solomon (approved):

running-recipes/getting-started.md
SimonVerhoeven: That looks good to me, it captures the essence and those pondering this will know where the bom would be used.

running-recipes/getting-started.md
```diff
@@ -298,6 +298,8 @@ After doing that, your `pom.xml` file should look similar to this:
 
 To double-check that everything is working, run the command `./mvnw rewrite:run`. Your project should be upgraded to Spring Boot 2 and all of the test classes should be updated to JUnit 5. Your `pom.xml` file will also have had its Spring dependencies updated, the JUnit 4 dependency removed, and the JUnit 5 dependency added.
 
+__note__ here [bill of materials (BOM) dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:bom\_import) is not mentioned for `dependencyManagement` as we cannot use it to version control the OpenRewrite plugin dependencies, unlike with the Gradle setup below.
```
mike-solomon, line 301:
```suggestion
{% hint style="info" %}
Maven does not currently support using a bill of materials (BOM) to specify plugin versions or dependencies. This means that you will have to specify the versions of each plugin by hand, unlike in the Gradle section below.
{% endhint %}
```

Issue #208, opened 2023-07-18
Add a list example to the method pattern docs


mike-solomon:
In the OpenRewrite Slack, it was brought up that there is some confusion about how to properly use method patterns + a List and that we should update the doc accordingly. 

Related thread: https://rewriteoss.slack.com/archives/C01A843MWG5/p1689618351249879

Related code: 

```java
    @Test
    void matcherForList() {
        rewriteRun(
          java(
            """
              import java.util.List;
              class Foo {
                  void bar(List<String> l) {}
                  void baz() {
                    bar(List.of("a", "b"));
                  }
              }
              """,
            spec -> spec.afterRecipe(cu -> assertThat(FindMethods.find(cu, "Foo bar(java.util.List)")).isNotEmpty())
          )
        );
    }
```

Olga suggested that this doc might be useful to point to: https://www.baeldung.com/spring-aop-pointcut-tutorial

Requested by: @timtebeek

PR #218, opened 2023-08-03
Update method patterns doc

mike-solomon:
To be less confusing and to provide more examples.

Fixes: https://github.com/openrewrite/rewrite-docs/issues/208

### DIFF
--- a/reference/method-patterns.md
+++ b/reference/method-patterns.md
@@ -4,54 +4,87 @@ description: A simple and powerful way of identifying method definitions and inv
 
 # Method patterns
 
-Many recipes operate on method declarations or invocations. Method patterns are a mechanism for configuring and developing such recipes. OpenRewrite's pointcut method expressions are derived from AspectJ. An OpenRewrite "method pattern" is comparable to an AspectJ's "Pointcut method expression".
+Traditionally, searching for or identifying methods was done by doing a simple text search for the name of the method or the parameters. You'd then look through all the results and ponder if this is _actually_ the method you wanted to find or not. When searching over a significant number of repositories, this becomes an impossible task, though. There could be hundreds of methods with the same name and it would be too much work to find or change just the ones you cared about.
+
+Fortunately, OpenRewrite offers a better alternative: method patterns. Method patterns provide a way to accurately and quickly identify one or more method definitions or invocations. These patterns can be used during the creation of recipes or by users when they run recipes.
+
+{% hint style="info" %}
+An OpenRewrite method pattern is comparable to an AspectJ ["pointcut expression"](https://www.baeldung.com/spring-aop-pointcut-tutorial).
+{% endhint %}
 
 ## Anatomy of a method pattern
 
-A pointcut method expression can identify one or more method definitions or invocations based on their:
+A method pattern can identify one or more method definitions or invocations based on:
 
-* Fully qualified receiver type
-* Method Name
-* Method Argument types
+* A fully qualified class name for where the method is defined or invoked (aka a fully qualified receiver type)
+* A method name
+* The method's argument types
 
-```text
-com.yourorg.ReceiverFullyQualifiedType methodName(argType1, argType2)
+For instance, let's say we have a class like this:
+
+```java
+package com.foo;
+
+class Bar {
+    void baz(String input1, int input2) {
+}
 ```
 
-There are two kinds of wildcard symbols:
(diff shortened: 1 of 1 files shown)


reference/method-patterns.md
```diff
@@ -4,54 +4,87 @@ description: A simple and powerful way of identifying method definitions and inv
...
-* `*` - Matches any one thing. Applicable to receiver type, method name, and arguments.
-* `..` - Matches zero or more. Applicable to receiver type and arguments.
+* `com.foo.Bar` - the fully qualified class name where the method is defined
+* `baz` - the method name itself
+* `String, int` – the method's argument types
+
+Combining the three would make this method pattern:
+
+```java
+com.foo.Bar baz(String, int)
```
timtebeek, line 42: Further down we use `org.Foo`; should we stay consistent to avoid confusion?

reference/method-patterns.md
```diff
@@ -4,54 +4,87 @@ description: A simple and powerful way of identifying method definitions and inv
...
+| `*..* foo(..)       `                       | Any method named `foo` which accepts any number of arguments. This is the equivalent of doing a generic text search for methods named `foo`.                                                                     |
+| `org..* foo(..)`                            | Any method named `foo` in a package that starts with `org`. This could be `orgmeow` or it could be `org.meow.bark` or so on.                                                                                     |
+| `org.openrewrite.java.* foo(..)`            | Any method named `foo` in a package that starts with `org.openrewrite.java.`. Note that the `.*` after `java` means that the method _must_ be in this package rather than some subsequent subpackages.           |
+| `org.Foo bar(java.util.List)`               | Exactly the `bar` method in the `org.Foo` package that takes in a single `java.util.List` as an argument.                                                                                                        |
```
timtebeek, line 81: With some of these I was wondering if this is actually how things work, such as the case matching `orgmeow`; did you find or test these somewhere? In the past we've had issues where our [docs were a little aspirational](https://rewriteoss.slack.com/archives/C01A843MWG5/p1686582393410069?thread_ts=1686563285.282649&cid=C01A843MWG5) rather than reflecting current afairs, and I'd like to avoid adding to the confusion.

reference/method-patterns.md
```diff
@@ -4,54 +4,87 @@ description: A simple and powerful way of identifying method definitions and inv
 
 # Method patterns
 
-Many recipes operate on method declarations or invocations. Method patterns are a mechanism for configuring and developing such recipes. OpenRewrite's pointcut method expressions are derived from AspectJ. An OpenRewrite "method pattern" is comparable to an AspectJ's "Pointcut method expression".
+Traditionally, searching for or identifying methods was done by doing a simple text search for the name of the method or the parameters. You'd then look through all the results and ponder if this is _actually_ the method you wanted to find or not. When searching over a significant number of repositories, this becomes an impossible task, though. There could be hundreds of methods with the same name and it would be too much work to find or change just the ones you cared about.
+
+Fortunately, OpenRewrite offers a better alternative: method patterns. Method patterns provide a way to accurately and quickly identify one or more method definitions or invocations. These patterns can be used during the creation of recipes or by users when they run recipes.
```
timtebeek, line 9: I'd maybe swap these two paragraphs around, as we sometimes link to the page, and then people want to quickly skim the first paragraph to see what this is about.

reference/method-patterns.md
mike-solomon: Jonathan wrote out examples here: https://moderneinc.slack.com/archives/C01S4M53THT/p1690921625049179?thread_ts=1690917553.347359&cid=C01S4M53THT

I was largely just taking his examples and converting them to this. It's possible that these don't work, though (as I never tested any of these).

I'll see if I can figure out how to test these to confirm.

reference/method-patterns.md
mike-solomon: Ended up re-writing the intro to be shorter. Hopefully that addresses your concern.

timtebeek: Thanks a lot!

reference/method-patterns.md
mike-solomon: @timtebeek FYI I can confirm that the `orgmeow` thing does indeed not work. I updated the docs to reflect that. `org..*` would match `org.meow` or `org.springframework` but _not_ `orgmeow`. I'm not entirely sure why but 🤷

Issue #170, opened 2023-05-22
Add paragraph about classpath on "Modifying methods with JavaTemplate"


timtebeek:
We have a page that describes how to use JavaTemplate for inserting new code:
https://docs.openrewrite.org/authoring-recipes/modifying-methods-with-javatemplate
This page currently does not mention classpath at all, which is often needed for recipes.

I propose we add a small section on how to configure the classpath for JavaTemplate, and then link to [how to use multiple versions](https://docs.openrewrite.org/authoring-recipes/multiple-versions) for more details.

mike-solomon:
On top of adding this, we'll also probably need to rewrite most of that doc as the JavaTemplate API has changed in rewrite 8.

mike-solomon:
I've updated the `JavaTemplate` page to work for rewrite 8 - but I wasn't sure what to document for classpath as I haven't done that before with templates. @timtebeek Is that something you're aware of and could provide an example for? No worries if not!

timtebeek:
Here's a [quick example](https://github.com/openrewrite/rewrite-testing-frameworks/blob/3c60d66b1aaf063e5ed0d862fa872b2a9cd90aa0/src/main/java/org/openrewrite/java/testing/hamcrest/AssertThatBooleanToAssertJ.java#L55) that uses `classPathFromResources`; which is slightly different in that we [support multiple versions](https://docs.openrewrite.org/authoring-recipes/multiple-versions), but the mechnism is roughly the same for regular dependencies as well, just using `.classpath(..)` instead. It's necessary to declare the classpath entries & imports whenever the template String uses methods that are from those dependencies rather than the JDK itself. Hope that helps! We can tackle any other questions via a quick chat if needed.

PR #220, opened 2023-08-08
Add a paragraph about classpath to JavaTemplate doc

mike-solomon:
As people will often need to do this.

Fixes: https://github.com/openrewrite/rewrite-docs/issues/170

@timtebeek Does this seem reasonable to you? Do you think more information should be added? Want to make sure this covers what you wanted in that issue ^_^

### DIFF
--- a/authoring-recipes/modifying-methods-with-javatemplate.md
+++ b/authoring-recipes/modifying-methods-with-javatemplate.md
@@ -273,6 +273,8 @@ public JavaIsoVisitor<ExecutionContext> getVisitor() {
 
 {% hint style="info" %}
 Note: When building a template, if you use a type that's not a base Java type, you will need to specify what package that type comes from. In our case, since we're adding a `Date` to the `MethodDeclaration`, we need to specify that this is a `java.util.Date` and not some other type of date. You can do that by adding an `imports` function with the packages for the types used in the template.
+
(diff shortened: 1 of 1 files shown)


authoring-recipes/modifying-methods-with-javatemplate.md
```diff
@@ -273,6 +273,8 @@ public JavaIsoVisitor<ExecutionContext> getVisitor() {
 
 {% hint style="info" %}
 Note: When building a template, if you use a type that's not a base Java type, you will need to specify what package that type comes from. In our case, since we're adding a `Date` to the `MethodDeclaration`, we need to specify that this is a `java.util.Date` and not some other type of date. You can do that by adding an `imports` function with the packages for the types used in the template.
+
+If your import is _not_ part of the JDK itself, you will need to specify a classpath so that the recipe will know where the function you're adding is coming from. You can do this via the `classpathFromResources` method such as in the [Migrate Hamcrest assertThat recipe](https://github.com/openrewrite/rewrite-testing-frameworks/blob/v2.0.9/src/main/java/org/openrewrite/java/testing/hamcrest/AssertThatBooleanToAssertJ.java#L55).
```
timtebeek, line 277: The `.classpathFromResources(...)` is slightly more advanced than the regular `.classpath(...)`, and only necessary when using multiple different versions of a dependency. I'd suggest here we use `.classpath(...)` with this example link: https://github.com/openrewrite/rewrite-static-analysis/blob/v1.0.4/src/test/java/org/openrewrite/staticanalysis/RemoveUnneededAssertionTest.java#L105
That allows users to use "regular" dependencies instead of the `recipeDependencies { parserClasspath(...) }` dependencies.

And then separately link to https://docs.openrewrite.org/authoring-recipes/multiple-versions for those who need more than `.classpath(...)` offers.

timtebeek (approved):

Issue #217, opened 2023-08-02
Update popular recipe guide to upgrade to Micronaut 4


timtebeek:
## What problem are you trying to solve?
A couple of pages still refer to upgrading to Micronaut 3, whereas Micronaut 4 was released on July 14th, with good migration support through helps from the Micronaut team.
https://github.com/openrewrite/rewrite-docs/blob/master/running-recipes/running-recipes.md
https://github.com/openrewrite/rewrite-docs/blob/master/running-recipes/popular-recipe-guides/migrate-to-micronaut-3-from-micronaut-2.md

## Describe the solution you'd like
Update these references to refer to https://docs.openrewrite.org/recipes/java/micronaut/micronaut3to4migration
Possibly incorporating some of the documentation steps seen at:
https://micronaut.io/2023/07/14/upgrade-to-micronaut-framework-4-with-openrewrite-and-gradle/
https://micronaut.io/2023/07/14/upgrade-to-micronaut-framework-4-with-openrewrite-and-maven/

mike-solomon:
Do you think it's worth making a new guide of going from 3 to 4 or can people use these same recipes to go from 2 to 4? I'm wondering if this is a situation where people would search for "Update to Micronaut 3" and then be confused when we don't have that.

Also - do you know where we could find examples of what these recipes do? In looking over the existing guide, we have some examples of how the code might change -- but I didn't see any examples in the Micronaut links. I'm not sure if those are still applicable for the Micronaut 4 upgrade or if they should be removed or something else.

timtebeek:
Probably a new guide only for 3 to 4 makes sense, since we don't currently chain the 2->3 migration into the 3->4 migration, and I'm not sure if that's safe to do so now. We can leave the 2 -> 3 migration guide as it is, and add a note at the top to both to cross reference them.

And perhaps a reference to this blog, and/or some of the information in there could help.
https://www.moderne.io/blog/micronaut-framework-4-0-automated-upgrade-with-openrewrite

mike-solomon:
Thanks for the clarification @timtebeek !

I think for this guide to be effective we'll need some samples of what this recipe changes. Do you know if there is a sample repository similar to the spring petclinic one where we could run this recipe and see what changes so I could get some real-world examples?

If not, are there any parts of the recipe that you think are especially important to call out? Perhaps we can use the tests to show a before and after for some of those? (see the below photo for options)

[...]

PR #222, opened 2023-08-10
Add Micronaut 3 -> 4 guide

mike-solomon:
I'm not familiar with Micronaut at all but I did my best to find some examples to add to this doc. Many of the examples are taken from the tests for the recipes that make up this larger recipe.

I chose to include the main ones that the Micronaut Guides needed to update in their migration.

Fixes: https://github.com/openrewrite/rewrite-docs/issues/217

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -141,6 +141,7 @@ protobuf
 quarkus
 Querydsl
 qute
+reactivestreams
 rebranded
 Recile
 recipespec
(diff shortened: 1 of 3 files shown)


timtebeek:
Looks good to me; thanks for adding this! Perhaps @jeremyg484 has something to add still (optional), but good to merge already.

mike-solomon:
Will go ahead and merge. Feel free to suggest changes after the fact and I'll add them later :)

PR #227, opened 2023-08-31
doc: polishing

SimonVerhoeven:
## What's changed?

Added some inline code blocks, and which => that

## What's your motivation?

To help out with the documentation

### Checklist


sambsnyd:
Thanks @SimonVerhoeven !

PR #229, opened 2023-09-02
doc: add Gradle Kotlin documentation

SimonVerhoeven:
## What's changed?

Also added the Gradle Kotlin configuration

## What's your motivation?

To enhance the documentation given Kotlin is the new default for Gradle

### Checklist


mike-solomon (approved):
Thanks for adding this! I think it’ll definitely help ^_^

PR #236, opened 2023-10-16
Use latest version of org.openrewrite.maven

kanish671:
The current version used is not available on mvnrepository.com We would like to update it to a valid version.

Screenshot 2023-10-16 at 3 25 16 PM

## What's changed?
Updated version of `rewrite-maven-plugin` dependency to an available version

### Checklist


timtebeek (approved):
Thanks!

mike-solomon (approved):
Thanks!

PR #239, opened 2023-10-23
doc: update comment

SimonVerhoeven:
## What's changed?

fixed a typo

### Checklist


timtebeek:
Thanks! Confusingly that line refers to a recipe that contains a misspelling, which we allow in the docs for that reason. Ideally I think we should rename that recipe and fix things properly that way, but that's for some other time.

PR #240, opened 2023-10-23
doc: add Develocity to the allowed word list

SimonVerhoeven:
## What's changed?

Added Develocity to the allowed word list

## What's your motivation?

Updated the current spell checker

It's maybe worth adding it to the expected file instead?

### Checklist


timtebeek:
Picked up in
- #239

PR #242, opened 2023-10-25
Add doc for types of recipes

mike-solomon:
To help people understand what the different types are. The intention is to add a more dedicated Refaster guide after this and then link to it from this doc.


authoring-recipes/types-of-recipes.md
```diff
@@ -0,0 +1,88 @@
...
+
+OpenRewrite offers support for three different types of recipes: declarative, refaster, and imperative. Each of these has its own advantages and disadvantages depending on the type of recipe you want to build. 
+
+In this doc, we'll briefly describe each of these so that you can determine what type of recipe works best for your needs. 
+
+## Declarative recipes
+
+[Declarative recipes](/reference/yaml-format-reference.md) are the simplest and most common recipes. They are entirely written in YAML, and they generally tie together existing recipes while adding some light configuration to them.
+
+The [commmon static analysis recipe](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/resources/META-INF/rewrite/common-static-analysis.yml) is a great example of this. It takes a bunch of simple recipes relating to static analysis and combines them into one larger recipe that can be run more easily. It does not add any logic, and it doesn't change the existing recipes.
```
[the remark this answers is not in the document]
mike-solomon: fixed

authoring-recipes/types-of-recipes.md
```diff
@@ -0,0 +1,88 @@
...
+
+{% hint style="success" %}
+As a best practice, if your recipe can be declarative (meaning it can be built out of other recipes), then you should make it declarative.
+{% endhint %}
+
+## Refaster templates
+
+Refaster templates are the "middle ground" of recipes. They offer more functionality than [declarative recipes](#declarative-recipes), but not as much as [imperative recipes](#imperative-recipes). On the other hand, compared to an imperative recipe, they're much quicker to create and require much less knowledge to get started.
+
+Refaster templates are ideal for straightforward replacements such as converting `StringUtils.equals(..)` to `Objects.equals(..)`. These are more than just a string replacement, though; they offer compiler and type support.
```
knutwannheden, line 40: I think we should try to make it clear that Refaster templates can only be used to replace one expression with another or one statement with another.

Also, we should link to the [Errorprone Refaster documentation](https://errorprone.info/docs/refaster), which isn't very exhaustive, but it still gives a nice overview.

Currently, we only support the `@BeforeTemplate` and `@AfterTemplate` annotations and our own form of the `@Matches` and `@NotMatches` annotations. The latter two aren't even documented in Refaster, except for in the [Javadoc](https://errorprone.info/api/latest/com/google/errorprone/refaster/annotation/package-summary.html). But since we plan to expand on this fairly soon, I don't think we really need to document it.

authoring-recipes/types-of-recipes.md
```diff
@@ -0,0 +1,88 @@
...
+  - org.openrewrite.staticanalysis.CatchClauseOnlyRethrows
+  - org.openrewrite.staticanalysis.ChainStringBuilderAppendCalls
+  # ...
+---
+```
+
+{% hint style="success" %}
+As a best practice, if your recipe can be declarative (meaning it can be built out of other recipes), then you should make it declarative.
+{% endhint %}
+
```
timtebeek, line 35: Perhaps it's good to add a second example here where we actually add some configuration to achieve common migration steps, like renaming a class, a method, package or dependency. These are often needed for both OSS libraries and frameworks, as well as company internal frameworks. An example can be found here:
https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml

authoring-recipes/types-of-recipes.md
```diff
@@ -0,0 +1,88 @@
...
+If your recipe can be defined via a declarative YAML file or via a refaster template, you should do so -- even if you can technically create an imperative recipe that does the same thing.
+
+The [FinalizeLocalVariables recipe](https://github.com/openrewrite/rewrite-static-analysis/blob/main/src/main/java/org/openrewrite/staticanalysis/FinalizeLocalVariables.java) is a great example of an imperative recipe.
+
+## Next steps
+
+* If you want more information about the types of recipes and the recipe execution pipeline, please see our [recipes documentation](/concepts-and-explanations/recipes.md)
+* If you want to write your own imperative recipe, please check out the [writing a Java refactoring recipe](/authoring-recipes/writing-a-java-refactoring-recipe.md) guide.
```
timtebeek, line 88: MIght be good to link to how to test recipes here as well, as that is a likely first step folks might want to take. Perhaps we should even spin that out into a separate doc, as it can be shared between declarative, refaster and imperative recipes.

mike-solomon: I added a link to the existing testing guide. We can potentially update that in the future to include more information on testing.

mike-solomon:
@timtebeek @knutwannheden I think I addressed your feedback. If you could take a look and let me know again when you have a sec, I'd appreciate it ^_^

timtebeek (approved):
Nice addition to our docs, thanks!

authoring-recipes/types-of-recipes.md
```diff
@@ -0,0 +1,128 @@
...
+  - org.openrewrite.staticanalysis.BigDecimalRoundingConstantsToEnums
+  - org.openrewrite.staticanalysis.BooleanChecksNotInverted
+  - org.openrewrite.staticanalysis.CaseInsensitiveComparisonsDoNotChangeCase
+  - org.openrewrite.staticanalysis.CatchClauseOnlyRethrows
+  - org.openrewrite.staticanalysis.ChainStringBuilderAppendCalls
+  # ...
+---
+```
+
+Another good example that demonstrates how you can use a declarative recipe to define a complex migration recipe is the [ApacheHttpClient migration recipe](https://github.com/openrewrite/rewrite-spring/blob/main/src/main/resources/META-INF/rewrite/apache-httpclient-5.yml):
```
timtebeek, line 32: Perhaps there's a good hook here to tell folks that they can compose such recipes through the recipe builder at https://app.moderne.io/recipes/builder ; that allows them to try it out, and has people explore the recipes in the platform.

mike-solomon: Great idea! Will add

Issue #189, opened 2023-06-29
Expand styles documentation


mike-solomon:
Right now, our [styles documentation](https://docs.openrewrite.org/concepts-explanations/styles) provides a relatively simple example but then does not really show how to use it. It links out to [another doc](https://docs.openrewrite.org/reference/yaml-format-reference#styles) that briefly mentions it. We should consider expanding on this to provide a fully working example of changing the style and applying it to a project.

[Related community Slack question](https://rewriteoss.slack.com/archives/C01A843MWG5/p1687980359411879)

mike-solomon:
Another instance of this being confusing to someone: https://rewriteoss.slack.com/archives/C01A843MWG5/p1698035295669299

PR #243, opened 2023-11-09
Redo styles doc

mike-solomon:
Fixes: https://github.com/openrewrite/rewrite-docs/issues/189

FYI @timtebeek - I'm going to merge this in as I know you're busy traveling. At some point if you wouldn't mind taking a look I'd appreciate it.


timtebeek:
I really like the rewrite you've done here! Should help to reduce questions around this topic.

My only suggestion would be to add a section at the bottom where we tell folks what recipes to use if they want to reformat their source code to conform to their configured style. I feel like that would be a common next step, as sometimes projects don't start out with a consistent style.

PR #245, opened 2023-11-10
Remove Running Rewrite without build tool plugins

timtebeek:
We now recommend the Moderne CLI as an alternative to Maven and Gradle plugins.


SUMMARY.md
```diff
@@ -1433,7 +1432,7 @@
       * [Ensure the GKE metadata server is enabled](reference/recipes/terraform/gcp/ensurethegkemetadataserverisenabled.md)
     * [Search](reference/recipes/terraform/search/README.md)
       * [Find Terraform resource](reference/recipes/terraform/search/findresource.md)
-  * [Core](reference/recipes/core-README.md)
+  * [Core](reference/recipes/core-readme.md)
```
mike-solomon, line 1435: Why are we changing the capitalization here? Was it incorrect before?

running-recipes/running-recipes.md
```diff
@@ -3,9 +3,8 @@
 * [Quickstart: Setting up your project and running recipes](/running-recipes/getting-started.md)
 * [Running Rewrite on a Gradle project without modifying the build](/running-recipes/running-rewrite-on-a-gradle-project-without-modifying-the-build.md)
 * [Running Rewrite on a Maven project without modifying the build](/running-recipes/running-rewrite-on-a-maven-project-without-modifying-the-build.md)
-* [Running Rewrite without build tool plugins](/running-recipes/running-rewrite-without-build-tool-plugins.md)
 * [Running Rewrite on a multi-module Maven project](/running-recipes/multi-module-maven.md)
-* [Popular recipe guides](/running-recipes/popular-recipe-guides/)
+* [Popular recipe guides](/running-recipes/popular-recipe-guides/README.md)
```
mike-solomon, line 7: Similar question here - was this link broken?

mike-solomon (approved):
Thanks for fixing this! Two minor questions below - feel free to merge after taking a look.

SUMMARY.md
timtebeek: My IDE gave me a warning on this link, and it does appear to be lower case:
https://github.com/openrewrite/rewrite-docs/blob/master/reference/recipes/core-readme.md
On other projects we try not to commit if the IDE has any warning, hence why I changed this now.

running-recipes/running-recipes.md
timtebeek: Similarly: IDE warning on a broken link; figured the result is likely the same, but one less warning this way if it's all the same to you.

SUMMARY.md
mike-solomon: Cool - thanks for taking care of both of those issues then!

PR #254, opened 2023-12-26
Document default `configLocation` in multi module Maven projects

lrkwz:
Changed docs as @timtebeek discovered


timtebeek:
Thanks a lot!

PR #258, opened 2024-01-18
Remove cycles from the docs

mike-solomon:


knutwannheden:
@mike-solomon We should revert this again. We decided that we still need to keep cycles for now.

mike-solomon:
changes have been reverted

PR #261, opened 2024-01-29
Fix path for library jars

gjoseph:
## What's changed?
Fix path for library jars in `src/main/resources/META-INF/rewrite/classpath` -- docs current suggest  `src/main/resources/META-INF/rewrite` which I found out was wrong (empirically + debugging)


timtebeek (approved):
You're absolutely right; thanks for providing the fix!

PR #275, opened 2024-05-17
Update runnertoextension.md

shalk:
### Checklist


timtebeek (approved):
Thanks for helping us improve the documentation example here! I've also [made a matching change to the recipe](https://github.com/openrewrite/rewrite-testing-frameworks/commit/f32fdb0e99aa34ebef73fcb46a6ef6a4c168e04c) where this page is generated from, such that it won't be overwritten going forward.

PR #282, opened 2024-06-16
Fix rule description markdown of tech.picnic.errorprone.refasterrules.TestNGToAssertJRulesRecipes

Philzen:
The description displayed at https://docs.openrewrite.org/recipes/tech/picnic/errorprone/refasterrules/testngtoassertjrulesrecipes is confusing, the mixture of HTML and markdown leads to disappearing relevant information such as this:

> (note that the type parameters of the map on the right-hand side will be inferred to be `` rather than ``).

Still clueless what is being referred to with "Some of the classes below " though.

Wondering if we couldn't cook up a nice recipe to do this cleanup automatically:
- remove leading and trailing underscore
- remove any `<p>` tags
- convert the `pre` block to markdown code block and remove the surrounding `{@code … }` custom tag

It would also be nice to have the Github-Link under "Recipe source" to link to the actual implementation.

### DIFF
--- a/reference/recipes/tech/picnic/errorprone/refasterrules/testngtoassertjrulesrecipes.md
+++ b/reference/recipes/tech/picnic/errorprone/refasterrules/testngtoassertjrulesrecipes.md
@@ -2,21 +2,24 @@
 
 **tech.picnic.errorprone.refasterrules.TestNGToAssertJRulesRecipes**
 
-_<p>Some of the classes below have TestNG `@BeforeTemplate`s that reference wildcard type
+Some of the classes below have TestNG `@BeforeTemplate`s that reference wildcard type
  bounds (`<?>`), while the associated AssertJ `@AfterTemplate`s reference stricter
  type bounds. This introduces the risk of producing invalid code. We do this anyway, because
  TestNG's wildcard types can cause javac to infer less specific types than AssertJ requires, while
  the appropriate (more specific) types _will_ be inferred properly when plugged into AssertJ's
  API.
 
- <p>The following is an example of a TestNG statement, which would not be rewritten if it weren't
+The following is an example of a TestNG statement, which would not be rewritten if it weren't
  for the wildcard matching (note that the type parameters of the map on the right-hand side will
  be inferred to be `<Object, Object>` rather than `<String, Object>`).
 
- <pre>{@code
+```java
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for the offer to help fix the description here! Unfortunately this one is a bit of a rabbit hole:

1. we generate our recipe documentation from the recipe jars themselves in rewrite-recipe-markdown-generator
2. this particular recipe is [found in rewrite-third-party](https://github.com/openrewrite/rewrite-third-party/blob/474b3461cdae4bcbd803d7efafd589903dddb969/build.gradle.kts#L56)
3. but only because that shades [error-prone-support](https://github.com/picnicsupermarket/error-prone-support)
4. which in turn [uses rewrite-templating](https://github.com/PicnicSupermarket/error-prone-support/blob/d81fe19836db9f0001fb9c01370a3b808c46b440/error-prone-contrib/pom.xml#L198-L200) to create recipes from templates
5. rewrite-templating can [extract the documentation either from JavaDoc or from annotations](https://github.com/openrewrite/rewrite-templating/blob/7eceffca8aee1e899b2614555dc938083538a20c/src/main/java/org/openrewrite/java/template/processor/RefasterTemplateProcessor.java#L434-L521)
6. In this particular case, the documentation likely comes from [the JavaDoc on TestNGToAssertJRules](https://github.com/PicnicSupermarket/error-prone-support/blob/d81fe19836db9f0001fb9c01370a3b808c46b440/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/TestNGToAssertJRules.java#L35-L51)

In this case I think we should likely update rewrite-templating to not put those wrapping `__` around a multi-line description, as a first formatting fix. Then once that's picked up in error-prone-support and released there, should the doc changes cascade here.

While we could merge this PR, the changes would otherwise be overwritten the next time we generate the documentation, which we tend to do every two weeks.

Philzen:
I had a hunch that would be a pointless PR due to some automated process generating this documentation. :wink: 

I've documented the issue here now: https://github.com/openrewrite/rewrite-templating/issues/93 (hope this is the correct repo to place this).

So i guess it doesn't matter much whether this is merged or closed, so i'll be fine with any.

timtebeek:
Thanks for logging that additional issue; I'll merge this for now until we can fix it upstream. 🙏🏻

PR #284, opened 2024-06-16
Fix xml syntax highlighting on "Maven plugin configuration" reference page

Philzen:
## What's changed?
See title.

## What's your motivation?
A less grey and cold looking maven config block in the documentation :wink:


timtebeek (approved):
Thanks a the quick fix!

PR #285, opened 2024-06-16
Replace hidden tabs with spaces in XML example on "Maven plugin configuration" reference page

Philzen:
One small followup to #284 

Apparently there are some tabs in the xml config example, which leads to it being displayed like this:

grafik


timtebeek (approved):
Thanks a lot! Seems like I didn't get notified of this one earlier; feel free to tag me if I missed something. I normally watch our Slack #github-integration channel, but it seems not everything appears there lately. 🤔

PR #286, opened 2024-06-18
Fix broken link to cursor-messaging section

Philzen:
There are two broken links in the middle paragraph of this section.

This PR fixes the first one.

### DIFF
--- a/authoring-recipes/recipe-conventions-and-best-practices.md
+++ b/authoring-recipes/recipe-conventions-and-best-practices.md
@@ -166,6 +166,7 @@ If you need to pass state between **different** visitors, then you should [overr
 ### Use cursor messaging instead of execution context messaging
 
 When passing state between functions in the same visitor, there are two main options: cursor messaging and execution context messaging.
+If you merely wish to extract some information from visiting downstream elements, then you can also [use a local visitor with a mutable parameter](/concepts-explanations/visitors#sharing-data-between-visitors).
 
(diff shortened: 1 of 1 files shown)


Philzen:
Of course, as Murphy's law dictates i discovered this while i was after for the information lurking beneath the other link :wink: 
Doing some git archeology suggests it used to point to the last section here: https://github.com/openrewrite/rewrite-docs/blob/4565919ccd1a8677834aeb9f887a263f71ecf503/authoring-recipes/recipe-conventions-and-best-practices.md#override-the-visit-method-if-you-need-pass-state-between-visitors

However i'm not sure this is still state-of-the-art as most of the examples use a "helper visitor" (btw, we should give that pattern a proper name) that has fields or methods that can then be queried from the consuming visitors like here: https://docs.openrewrite.org/concepts-explanations/visitors#sharing-data-between-visitors

So that link needs to be fixed as well, if you advise how i can add that to this PR, but i'm happy for anybody to do it (and looking forward to the correct answer).

timtebeek (approved):
Thanks for fixing the broken link & asking for clarification on the recommended patterns. I've added a line that  hope helps, but I'm open to improvements still. Indeed would be good to give that pattern a more easily discoverable name.

Philzen:
@timtebeek Thanks! 

Not sure if there was a little misunderstanding (i admit i sometimes write too much so the gist gets lost in the noise :see_no_evil:)  … but the broken link in question that i was referring to is still there:

grafik

https://github.com/Philzen/rewrite-docs/blob/b6691b7dd3b4253375f8173d82d3f8aec55c714c/authoring-recipes/recipe-conventions-and-best-practices.md?plain=1#L164

timtebeek:
Thanks for pointing that out! Can be hard to know exactly what I missed when jumping between issues. I've pushed a new change (https://github.com/openrewrite/rewrite-docs/commit/a777a28c80030f3ccbeb3fd1527161c0bd43f4c5) that hopefully documents the correct way forward, based on how we've changed recipes linked in the former docs that you so helpfully linked. :pray:

PR #292, opened 2024-07-22
Document Gradle data tables

timtebeek:
Following the addition of data tables in
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/310


mike-solomon (approved):
Thanks!

mike-solomon:
> @mike-solomon We'll similarly want to update the recipe markdown generator to add that `exportDatatables = true` for recipes that produce data tables. Is that something you're ok taking on?

To double-check that I understand, you're specifically asking for the `Gradle` example to include `exportDatatables = true` in the `build.gradle` file, correct? Or is there more that you'd like to see automated, too?

PR #299, opened 2024-09-10
Update addplugindependency.md - missing comma

janmaterne:
Just fixed a typo

### Checklist


timtebeek:
Thanks a lot for pointing this out @janmaterne ; I've picked this up in https://github.com/openrewrite/rewrite-recipe-markdown-generator/commit/fa7e12edd71c6030ad713a2e523692d9182058ce as we generate that part of the docs. That also means all 2700+ recipe pages should be improved based on your feedback, so thanks a lot! :)

PR #303, opened 2024-09-26
Add .NET recipes

bryceatmoderne:
### Checklist


mike-solomon:
How were these docs generated? Is there a reason we need them now vs. waiting until the artifact is available?

Also - in order for any of these to appear, you'll need to update the [SUMMARY.md file](https://github.com/openrewrite/rewrite-docs/blob/master/SUMMARY.md)

bryceatmoderne:
> How were these docs generated? Is there a reason we need them now vs. waiting until the artifact is available?
> 
> Also - in order for any of these to appear, you'll need to update the [SUMMARY.md file](https://github.com/openrewrite/rewrite-docs/blob/master/SUMMARY.md)

These changes aren't urgent. I generated them using `rewrite-recipe-markdown-generator` using a snapshot version of `rewrite-dotnet` since I had not added it to `rewrite-bom` until recently. 

Thanks for the `SUMMARY.md` reminder. I'll update the diff.

mike-solomon:
I've added these docs via a separate PR. I was worried about pushing the docs over the edge again so I deleted a bunch of recipes from the docs for now and added this in - along with the generated docs from it. Thanks!

Issue #308, opened 2024-10-15
Links that yield a 404


nielsdebruin:
## What problem are you trying to solve?

While browsing the documentation came across some links that yielded a 404. To get a better picture of this problem I wrote a script to find all URLs that point to either github.com or doc.openwrite.com. The reasoning is that GitHub links, which generally reference code should always be valid aswell as internal documentation link. Below you will find the following output of the script, in the format of `{URL}, {FILE}`.

## Sidenote
Perhaps an open write recipe which actually checks this could be a nice addition to the project in general, or for this project a Github action that e.g. runs weekly and simply excute a script similar to the one I wrote.  

```
URL, FILE
https://docs.openrewrite.org/concepts-explanations/styles, rewrite-docs/docs/recipes/java/orderimports.md
https://github.com/openrewrite/rewrite-java-test, rewrite-docs/docs/reference/latest-versions-of-every-openrewrite-module.md
https://github.com/openrewrite/rewrite-java-test/releases/tag/v8.37.1, rewrite-docs/docs/reference/latest-versions-of-every-openrewrite-module.md
https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime, rewrite-docs/docs/recipes/java/migrate/javax/addjaxwsruntime.md
https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime, rewrite-docs/docs/recipes/java/migrate/javax/addjaxwsruntime$addjaxwsruntimegradle.md
https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/javax/AddJaxwsRuntime, rewrite-docs/docs/recipes/java/migrate/javax/addjaxwsruntime$addjaxwsruntimemaven.md
https://github.com/openrewrite/rewrite-migrate-java/blob/main/src/main/java/org/openrewrite/java/migrate/util/OptionalStream.java, rewrite-docs/docs/recipes/java/migrate/util/optionalstreamrecipe.md
https://github.com/openrewrite/rewrite-protobuf, rewrite-docs/docs/reference/latest-versions-of-every-openrewrite-module.md
https://github.com/openrewrite/rewrite-protobuf/releases/tag/v8.37.1, rewrite-docs/docs/reference/latest-versions-of-every-openrewrite-module.md
[...]

PR #309, opened 2024-10-15
Fix link

nielsdebruin:
## What's changed?
Fixed a documentation link based on #308

### DIFF
--- a/docs/recipes/java/orderimports.md
+++ b/docs/recipes/java/orderimports.md
@@ -9,7 +9,7 @@ import TabItem from '@theme/TabItem';
 
 **org.openrewrite.java.OrderImports**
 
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks for spotting this one @nielsdebruin ! Anything under `/recipes/` is generated from the recipe source, so I've fixed this in https://github.com/openrewrite/rewrite/commit/e88596ec48a47cdb096287b8b8edff9992c0bec3

We're considering removing those files from the Git repository such that there's no such confusion. We already have (had?) an automation to point this out on PRs. /cc @mike-solomon does that still work?

mike-solomon:
@timtebeek I thought I updated it to work but I guess it didn't work as I expected: https://github.com/openrewrite/rewrite-docs/blob/master/.github/workflows/unneeded_recipe_updates.yml

I'll create an issue for this to look into it.

PR #312, opened 2024-10-24
Update bigdecimalrulesrecipes$bigdecimalvalueofrecipe.md

al1lhomme:
add tag referencing sonar rule

## What's changed?
 i have add a tag section to reference the corresponding sonar rule 

## What's your motivation?
i was first searching this recipe by sonar rule reference, i dont find it and try other keywords

format is copied from https://docs.openrewrite.org/recipes/staticanalysis/writeoctalvaluesasdecimal#tags

### DIFF
--- a/docs/recipes/tech/picnic/errorprone/refasterrules/bigdecimalrulesrecipes$bigdecimalvalueofrecipe.md
+++ b/docs/recipes/tech/picnic/errorprone/refasterrules/bigdecimalrulesrecipes$bigdecimalvalueofrecipe.md
@@ -30,6 +30,10 @@ static final class BigDecimalValueOf {
 ```
 
 
+### Tags
+
+* [RSPEC-S2111](https://sonarsource.github.io/rspec/#/rspec/S2111)
+
 ## Recipe source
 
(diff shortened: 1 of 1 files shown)


timtebeek:
hi @al1lhomme ! Thanks for looking to improve the docs. There's a a few indirections here:
1. our docs are generated based on tags defined on recipes, so a change here would be overwritten next time we generate the docs
2. these recipes are generated based on Refaster rules, which do support documentation, but don't yet support tags specifically
3. the refaster rules that serve as input for these recipes are maintained by our friends at Picnic: https://error-prone.picnic.tech/refasterrules/BigDecimalRules/#bigdecimalvalueof

That means your change here would only temporarily feature in the docs if merged. But if you [change the JavaDoc here](https://github.com/PicnicSupermarket/error-prone-support/blob/9b4cbfb84dcebd9dbdd2235ca2b76675dfcb8f2e/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java#L54-L58) to add a reference to that tag, then it should show after the next release makes it in here. I hope that helps!

timtebeek:
Although interestingly there's also this Suppress warnings line on their template:
https://github.com/PicnicSupermarket/error-prone-support/blob/9b4cbfb84dcebd9dbdd2235ca2b76675dfcb8f2e/error-prone-contrib/src/main/java/tech/picnic/errorprone/refasterrules/BigDecimalRules.java#L60

Perhaps we odd to pull that into our tags to show those in the docs; neat! Logged as a suggestion here:
- https://github.com/openrewrite/rewrite-templating/issues/113

PR #318, opened 2024-11-05
NOISSUE: Fix typo in JUnit4 to JUnit5 recipe

varunu28:
## What's changed?
Small fix in the typo in the maven part of the recipe https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5
The `build` closing tag does not have the correct format & is missing the `/`
image

## What's your motivation?
I am using this recipe for migrating a bunch of modules & copying the recipe from the clipboard & pasting it the `pom.xml` leads to syntax error.


timtebeek:
Thanks @varunu28 ! Very kind to immediately fix this. :) Must help folks that copy and paste this snippet.

PR #326, opened 2024-11-26
Add a default constructor for test only to pass testing in Java recipe guide.

addozhang:
## What's changed?

```
java.lang.AssertionError: [Recipe must be able to instantiate via RecipeIntrospectionUtils] 
Expecting code not to raise a throwable but caught
  "org.openrewrite.config.RecipeIntrospectionException: Unable to call primary constructor for Recipe class ...
...
```

I found this is issue with [this commit](https://github.com/openrewrite/rewrite/commit/0eaa9a3542850578db3bf92d633b43863faf7f7f#diff-bd1b934e99c7cf3526b38925eb633f2ad10afbc2a2275d5ed0c908c393329542R168). In this commit, The `RewriteTest` will invoke `RecipeIntrospectionUtils.constructRecipe(recipe.getClass())` to do some validation. This requires our recipe should contain a default constructor.

Add a default constructor will fix it.
    
```java
    @TestOnly
    public SayHelloRecipe() {
        fullyQualifiedClassName = "";
    }    
```

Not sure.

### Checklist

### DIFF
--- a/docs/authoring-recipes/writing-a-java-refactoring-recipe.md
+++ b/docs/authoring-recipes/writing-a-java-refactoring-recipe.md
@@ -56,6 +56,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.EqualsAndHashCode;
 import lombok.Value;
+import org.jetbrains.annotations.TestOnly;
 import org.jspecify.annotations.NonNull;
 import org.openrewrite.*;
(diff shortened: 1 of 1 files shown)


timtebeek:
Thanks a lot for the fix here @addozhang , also for pulling in that context of where it now fails. I've removed the `@TestOnly` annotation as we want to minimize what folks need to have available, as we similarly stripped Lombok from the test examples as well. With that we're good to merge. ❤️

PR #344, opened 2025-01-22
Redesign 2024 updated

bookwyrm:
Redesign for docs site including all feedback.


mike-solomon (approved):
Thank you for the rework!

PR #346, opened 2025-01-24
refactor: Common static analysis issues

sambsnyd:
Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.staticanalysis.CommonStaticAnalysis?organizationId=T3BlblJld3JpdGU%3D

### DIFF
--- a/src/plugins/replace-tokens.js
+++ b/src/plugins/replace-tokens.js
@@ -11,7 +11,9 @@ function remarkTokenReplacer(options) {
       (token) => node.value && node.value.includes(token)
     );
 
-    if (!hasToken) return
+    if (!hasToken) {
+        return
+    }
 
     Object.entries(replacements).forEach(([token, value]) => {
       if (!node.value.includes(token)) {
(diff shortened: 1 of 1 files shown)


timtebeek (approved):

src/plugins/replace-tokens.js
```diff
@@ -58,7 +60,9 @@ function remarkTokenReplacer(options) {
       (token) => node.url && node.url.includes(token)
     );
 
-    if (!hasToken) return;
+    if (!hasToken) {
+        return
+    };
```
timtebeek, line 65: Moving the semicolon here is a bit suspect.

knutwannheden: I suspect it has to do with the Semicolon marker. We should modify this recipe to make it aware of that.

sambsnyd:
I opened this to record footage of issuing PRs for a video.
Agree we should fix whatever was going on with that semicolon

PR #352, opened 2025-02-21
Update getting-started.md

johan-kragt:
## What's changed?
I removed the requirement on JDK 11, and changed it to JDK 11 +. I'm not sure how this relates to the Gradle JVM.

## What's your motivation?
I tested the Getting Started with JDK 21, and it seems to work fine.


timtebeek (approved):
Thanks for the immediate fix!

PR #353, opened 2025-02-21
Update recipe-development-environment.md

johan-kragt:
## What's changed?
I changed the mvn generate command to use archetype version 1.5 and made the appropriate changes to the pom versions. This way, a project will be generated using Unit 5 instead of 4. I also provided a single version of lombok in dependency management.

## What's your motivation?
Using archetype version 1.4 actually gave some errors while building the project.


timtebeek (approved):
Thanks!

PR #355, opened 2025-02-21
Update modifying-methods-with-javatemplate.md

johan-kragt:
## What's changed?
Fixed assignment of the addMethodBodyTemplate example (pre maybeAutoFormat).

## What's your motivation?
Because the assignment is missing, the follow up text describes a situation that doesn't occur.


mike-solomon (approved):

mike-solomon:
Thanks for the find and fix!

PR #357, opened 2025-02-28
Update method-patterns.md

CodexRaunak:
Update the `ChangeMethodNameVisitor` class.

## What's changed?
- The `getType()` returns a JavaType, changed it to `getMethodType()`. 
- With `withType()` we have to mention the `withMethodType()` as well or it will give the error of missing type or invalid type error.

## What's your motivation?
I was making a recipe to change the method name, when I refered docs I used the code provided in the `ChangeMethodNameVisitor `but it errored to `missing type or invalid type in LST`. Then I made the required changes, maybe this was the correct one used before. But now the current change works.

I reffered this https://github.com/openrewrite/rewrite/blob/main/rewrite-java/src/main/java/org/openrewrite/java/ChangeMethodName.java#L103-L110

**Before:**
```
JavaType.Method type = m.getType();
            if (type != null) {
                type = type.withName(newMethodName);
            }

            m = m.withName(m.getName().withName(newMethodName))
                    .withType(type);
        }

```
        
image
Or also change made in 2 cycles error.

**After:**
No error, recipe works fine.
### Checklist

### DIFF
--- a/docs/reference/method-patterns.md
+++ b/docs/reference/method-patterns.md
@@ -182,7 +182,7 @@ class ChangeMethodNameVisitor extends JavaIsoVisitor<ExecutionContext> {
 
         // The enclosing class of a J.MethodDeclaration must be known for a MethodMatcher to match it
         if (methodMatcher.matches(method, classDecl)) {
-            JavaType.Method type = m.getType();
+            JavaType.Method type = m.getMethodType();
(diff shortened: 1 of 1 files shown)


mike-solomon (approved):
Thanks so much for catching and fixing this!

PR #358, opened 2025-03-05
Add example of exclusion configuration for rewrite maven plugin

dralagen:
## What's changed?
Add exemple on exclusion parameters

## What's your motivation?
the name of configuration is "exclusions" but in arg command is `rewrite.exclusions` 
I found an ambiguity with config "skipMavenParsing" and exemple is `-DskipMavenParsing=true`

### Checklist


mike-solomon (approved):
Thanks for the callout + fix + explanation. I agree that is a bit confusing!

PR #359, opened 2025-03-05
Remove parallel suggestion from jar recipe install command

Laurens-W:
## What's changed?

Remove `--parallel` suggestion from the `mod config recipes jar install` command as it causes failures due to locking

### Checklist


mike-solomon (approved):
This is an automated doc - so the automation in the markdown generator will also need to be updated to stop this from happening. I can take care of that, though.

PR #361, opened 2025-03-06
Explain using jakarta.annotation.Generated for refaster recipes

bmuschko:
## What's changed?

An extension to the Refaster template recipe docs that explains the handling of the `Generated` annotation.

## What's your motivation?

Refaster template recipe projects that use Java 9+ need to switch to a different annotation than `javax.annotation.Generated`. While you can find instructions on doing this in different PRs and issues, the process of equipping a project with the necessary code isn't explained explicitly. The added docs section explains it.

No

@timtebeek

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -4,3 +4,4 @@ https
 ssh
 ubuntu
 workarounds
+Arewrite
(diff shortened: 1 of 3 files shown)


timtebeek (approved):
Thanks a lot for making this easier to find for the next person to stumble across this.

PR #363, opened 2025-03-10
Add a new FAQ about ignoring formatting or types

mike-solomon:
Based on this Slack question: https://rewriteoss.slack.com/archives/C01A843MWG5/p1741425521114579

"Hi, is there a way to tell rewriteTest to ignore the formatting when comparing before and after java? Or ideally to compare AST is equivalent, ignoring if static imports or fully qualified types are used."


timtebeek (approved):
We can add a last sentence saying we recommend fixing the underlying causes of those differences, although that might be obvious.

mike-solomon:
Thanks for the review and feedback @timtebeek !

Issue #202, opened 2023-07-11
Overhaul JavaTemplate documentation


knutwannheden:
As of OpenRewrite 8 the JavaTemplate API changed quite a bit and also introduces the distinction of context-free and -sensitive templates. This calls for an overhaul of the documentation on this topic.

mike-solomon:
This issue requires the help of someone on the dev team to read through the [JavaTemplate doc](https://docs.openrewrite.org/concepts-explanations/javatemplate) and then provide information on what needs to change. Tagging @sambsnyd for potential prioritization / help in the future. Marking this as blocked for now.

joanvr:
I'm going to collect in this comment the bits that I found during support or using the documentation that needs changes:
https://github.com/openrewrite/rewrite-docs/blob/0f262f5c3141e53dad1f50ad981e47a232546965/concepts-and-explanations/javatemplate.md?plain=1#L121

mike-solomon:
As part of this - we should make sure to include information about context sensitivity. [Related slack post](https://moderneinc.slack.com/archives/C04HR6EJ38D/p1692724445796149?thread_ts=1692722704.781429&cid=C04HR6EJ38D)

knutwannheden:
I checked the l[atest documentation](https://docs.openrewrite.org/concepts-and-explanations/javatemplate) and could not find any reference to context-free or -sensitive. I can help you with this.

mike-solomon:
Cool thank you! Feel free to open a PR or shoot me a message with the information or whatever else is easiest for you.

mike-solomon:
Closing this as it should've been addressed: https://github.com/openrewrite/rewrite-docs/commit/2f520bbf3ed7b76ef74277e5cd865e0d7e1d9d59

PR #364, opened 2025-03-10
Add context-free vs. context-sensitive to JavaTemplates

mike-solomon:
To address: https://github.com/openrewrite/rewrite-docs/issues/202

### DIFF
--- a/docs/concepts-and-explanations/javatemplate.md
+++ b/docs/concepts-and-explanations/javatemplate.md
@@ -120,6 +120,33 @@ JavaTemplate.builder("Duration.ofMillis(#{any(int)})")
 There are many examples of `JavaTemplate` stubs in [rewrite-testing-frameworks](https://github.com/openrewrite/rewrite-testing-frameworks) and [rewrite-spring](https://github.com/openrewrite/rewrite-spring).
 :::
 
+## Context-free vs. Context-sensitive JavaTemplates
+
+There are two types of JavaTemplates – each having their own advantages/disadvantages:
+
+A **context-free** JavaTemplate does not depend on the surrounding LST. It is a static template that simply replaces or inserts code without considering the broader context of the source file. This is particularly useful for simple transformations like replacing method calls or inserting boilerplate code.
+
+For instance, the following template, when applied, would not check the existing imports or variables in scope:
+
+```java
+JavaTemplate.builder("System.out.println(\"Hello, World!\");")
+            .build();
+```
+
+The main advantage of context-free templates is that they only need to be parsed once before their parameters can be substituted. This has a much smaller performance impact on recipe runs.
+
+A **context-sensitive** JavaTemplate, on the other hand, is aware of and adapts to the surrounding LST. It can capture elements from the existing code, such as variable names, types, or method parameters.
+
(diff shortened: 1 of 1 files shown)


docs/concepts-and-explanations/javatemplate.md
```diff
@@ -120,6 +120,51 @@ JavaTemplate.builder("Duration.ofMillis(#{any(int)})")
...
+public void logMessage() {
+    String message = "Application started";
+    System.out.println(message);
+}
+```
+
+We want to change the `System.out.println` to be a logging method, but we don't want to use the variable `message`. We could write a template like:
+
+```java
+JavaTemplate.builder("logger.info(#{any(String)});")
```
knutwannheden, line 152: This template does actually not have to be declared as context-sensitive. That is because also context-free templates can have parameters, like the `#{any(String)}` in this example. The context-sensitive template for this example would have been: `"logger.info(message);"`. Maybe that example is however a bit too simplistic?

docs/concepts-and-explanations/javatemplate.md
```diff
@@ -120,6 +120,51 @@ JavaTemplate.builder("Duration.ofMillis(#{any(int)})")
...
+When applied, the template might generate:
+
+```java
+public void logMessage() {
+    String message = "Application started";
+    logger.info(message);
+}
+```
+
+This context of the surrounding code comes with a performance impact, though. That's not to say that you _shouldn't_ use them but, rather, that you should be aware of the cost of using them.
```
knutwannheden, line 166: Maybe we can here also explain what that cost is: Every time a context-sensitive JavaTemplate is applied, a small Java source file capturing all imports and declarations that are in scope needs to be generated and compiled using the Java compiler.

docs/concepts-and-explanations/javatemplate.md
mike-solomon: Hmm I get what you're saying - but ya I want to demonstrate the substitution. Do you have any ideas for a better example?

As I'm not as involved in recipes as you are, I'm happy to defer to your expertise here - whether that be putting it as just `"logger.info(message);"` or finding a better example.

docs/concepts-and-explanations/javatemplate.md
mike-solomon: Great suggestion! Updated.

docs/concepts-and-explanations/javatemplate.md
timtebeek: No immediate better examples come to mind; indeed when referencing a field and it's somehow awkward to get a type attributed reference to insert into the template is when contextSensitive makes sense.

docs/concepts-and-explanations/javatemplate.md
```diff
@@ -120,6 +120,43 @@
...
+                .imports("java.nio.charset.StandardCharsets")
+                .build().apply(
+                        getCursor(),
+                        m.getCoordinates().replaceArguments(),
+                        m.getArguments().toArray());
+```
+
+This has to be context-sensitive because, for the type attribution, it needs to be generated as part of a call to the correct method.
+
+In other words, context sensitivity makes the most sense when you want to refeference a field that's awkward/difficult to get a type attributed reference of.
```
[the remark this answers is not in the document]
mike-solomon: fixed

PR #367, opened 2025-03-17
Yaml interpretation info

Laurens-W:
## What's changed?

Add info regarding YAML interpretation to avoid confusion

### Checklist

### DIFF
--- a/docs/reference/yaml-format-reference.md
+++ b/docs/reference/yaml-format-reference.md
@@ -9,7 +9,7 @@ OpenRewrite allows you to create [recipes](../concepts-and-explanations/recipes.
 To help you confidently define recipes and styles in YAML, this guide will walk you through all of the ways you can configure an OpenRewrite YAML file.
 
 :::info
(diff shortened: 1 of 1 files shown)


docs/reference/yaml-format-reference.md
```diff
@@ -8,6 +8,10 @@ OpenRewrite allows you to create [recipes](../concepts-and-explanations/recipes.
 
 To help you confidently define recipes and styles in YAML, this guide will walk you through all of the ways you can configure an OpenRewrite YAML file.
 
+:::info
+Note that values passed to declarative recipes are subject to YAML interpretation. So `1.20` would be interpreted as a float, causing the trailing 0 to be dropped.
```
timtebeek, line 12: Perhaps it helps to add a small suggestion of what to do instead; does wrapping in double quotes help for instance to make it a string? Or use `1.20.x`?

timtebeek (approved):

PR #379, opened 2025-04-18
Update simplifyassertjassertion.md

stefanodallapalma:
Fixed typo

### Checklist


greg-at-moderne:
Thanks for spotting this.
If you want to fix this, you need to fix it here:
https://github.com/openrewrite/rewrite-testing-frameworks/blob/d4f5d15002a365b22df02d316d46c8be54dacfb7/src/main/java/org/openrewrite/java/testing/assertj/SimplifyAssertJAssertion.java#L67

PR #383, opened 2025-04-26
Scanning recipes can now be used as preconditions

timtebeek:
- After https://github.com/openrewrite/rewrite/pull/5340


docs/reference/yaml-format-reference.md
```diff
@@ -52,7 +52,7 @@ You can find the full recipe schema [here](https://github.com/openrewrite/rewrit
 
 ### Preconditions
 
-Preconditions are used to limit which source files a recipe is run on. This is commonly used to target specific files or directories, but any recipe which is not a `ScanningRecipe` can be used as a precondition. 
+Preconditions are used to limit which source files a recipe is run on. This is commonly used to target specific files or directories, but any recipe can be used as a precondition. 
 
 :::info
 Preconditions are a **per-file** check. If a file passes the precondition check, all recipes will be run on it.
```
timtebeek, line 58: There's a sentence below here that says
> If you need to check if **your repository** meets certain criteria, instead (e.g., ensuring that a test source set exists), then you will need to write a custom `ScanningRecipe`.

What's the rephrased recommendation now that we can use scanning recipes as preconditions? Or should we just remove that line and potentially adjust this line?

mike-solomon (commented):
The changes LGTM -- although I don't have an answer to your question and it sounds like Olga is saying that Scanning recipes can only be used as preconditions for _declarative_ recipes instead of imperative? If so, we should perhaps leave the text in about them not being available as preconditions but tweak it to say it can't be used as a precondition for an imperative recipe?

okundzich:
My only comment was that this precondition is available for declarative recipes not for imperative.

docs/reference/yaml-format-reference.md
MBoegers: I would love to keep this note somehow, it answers expected questions from the beginning. We could merge it into the preconditions section like:

```markdown
### Preconditions

Preconditions are used to limit which source files a recipe is run on. 
This is commonly used to target specific files or directories.
Any recipe can be a precondition, to implement a custom **repository wide** preconditions (e.g., ensuring that a test source set exists) a `ScanningRecipe` is the best choice.
```

sambsnyd:
I've added a commit clarifying things. Take a look and let me know what you think.

docs/reference/yaml-format-reference.md
```diff
@@ -123,6 +121,50 @@ recipeList:
...
+type: specs.openrewrite.org/v1beta/recipe
+name: org.sample.FlawedSonarStaticAnalysis
+displayName: Fix sonar issues 
+description: >- 
+  This recipe is intended to apply common static analysis issues only to gradle projects that apply the sonar plugin. 
+  In practice this precondition will limit changes to apply ONLY to build.gradle(.kts) files. 
+  But common static analysis issues are found within the sources, not the build scripts, so no changes will ever be made.
+preconditions:
+  - org.openrewrite.gradle.search.FindPlugins:
+      pluginId: org.sonarqube
```
timtebeek, line 139: Perhaps @mike-solomon will have some ideas around this ; I'm usually hesitant to add "don't do this" into docs, as then folks happily copy paste the thing they shouldn't do and complain. I'd lead with or only show the thing they should do, and explain that not all recipes are a good fit separately in a warning block.

mike-solomon: I agree with your sentiment. Will rewrite this section a bit to hopefully get the same meaning across.

timtebeek (commented):
Thanks for the help all! I think we can merge this iteration already.

PR #388, opened 2025-05-19
Rewrite multiple versions of a library doc

mike-solomon:
To include new Maven example, reorder some things, and make stuff generally clearer.

- Fixes: https://github.com/moderneinc/customer-requests/issues/1040

@timtebeek If you could double-check that the edits are correct and address your request I'd appreciate it.

### DIFF
--- a/docs/authoring-recipes/multiple-versions.md
+++ b/docs/authoring-recipes/multiple-versions.md
@@ -44,22 +44,18 @@ The benefit of a TypeTable is that you don't need to pull down the entire JAR to
 
 #### Using TypeTables in your tests
 
-With the TypeTable file created, you can use the `classpathFromResources` function in your tests. For instance, if you want to default to using `junit-4.13.2` and `mockito-core-3.12.4` in your tests, you could add this to your test file:
+With the TypeTable file created, you can use the `classpath(String...)` function in your tests. For instance, if you want to default to using `junit-4.13.2` and `mockito-core-3.12.4` in your tests, you could add this to your test file:
 
 ```java
 @Override
 public void defaults(RecipeSpec spec) {
     spec
         .parser(JavaParser.fromJavaVersion()
-            .classpathFromResources(new InMemoryExecutionContext(), "junit-4.13.2", "mockito-core-3.12.4"));
+            .classpath("junit-4.13.2", "mockito-core-3.12.4"));
 }
 ```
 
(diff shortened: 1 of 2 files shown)


docs/authoring-recipes/multiple-versions.md
```diff
@@ -57,20 +34,103 @@ recipeDependencies {
...
 
 The benefit of a TypeTable is that you don't need to pull down the entire JAR to be able to compile specific types. As this TypeTable is not a class, and it can't be re-hydrated into anything executable, it does not trigger security vulnerability scanners ([which was a concern with the previous recipeDependencies task](../reference/faq.md#why-do-artifact-scanners-detect-vulnerabilities-in-recipe-artifactsjars) – more on that below). Furthermore, it greatly reduces the size of your dependencies (by roughly 90%), which makes releases significantly smaller.
 
-With the TypeTable file created, you can use the `classpathFromResources` function [in the same way as above](#manually-copying-jars-and-using-the-classpathfromresources-function).
+#### Using TypeTables in your tests
+
+With the TypeTable file created, you can use the `classpathFromResources` function in your tests. For instance, if you want to default to using `junit-4.13.2` and `mockito-core-3.12.4` in your tests, you could add this to your test file:
```
timtebeek, line 47: Generally speaking we recommend folks use `.classpath(String...)` in tests, and `.classpathFromResources(ExecutionContext, String...)` only in recipes, unless those are exactly the same. That helps the .tsv.zip file remain  as small as possible for lightweight recipe jars.

docs/authoring-recipes/multiple-versions.md
```diff
@@ -44,22 +44,18 @@ The benefit of a TypeTable is that you don't need to pull down the entire JAR to
 
 #### Using TypeTables in your tests
 
-With the TypeTable file created, you can use the `classpathFromResources` function in your tests. For instance, if you want to default to using `junit-4.13.2` and `mockito-core-3.12.4` in your tests, you could add this to your test file:
+With the TypeTable file created, you can use the `classpath(String...)` function in your tests. For instance, if you want to default to using `junit-4.13.2` and `mockito-core-3.12.4` in your tests, you could add this to your test file:
```
timtebeek, line 47: I feel there's some potential for confusion here: folks do not need type tables to use `classpath(String...)`; perhaps best to explain type tables only in the context of using JavaTemplate and JavaParser in a recipe, as that's what it's intended for, and remove the references to testing here.

mike-solomon: 👍 alright removed this section.

PR #391, opened 2025-06-08
Indicate previously generated files are seen by scanner in a ScanningRecipe

timtebeek:
I recall @sambsnyd adding this little implementation detail to a thread discussing this, and it seems supported by the implementation:
https://github.com/openrewrite/rewrite/blob/0a2f1551826861d29c5a2297d81174d49fec6254/rewrite-core/src/main/java/org/openrewrite/scheduling/RecipeRunCycle.java#L125-L169


mike-solomon (approved):
Thanks!

PR #399, opened 2025-07-09
Add "Semantics class shortcut" section to JavaTemplate

jevanlingen:
## What's changed?
A new "Semantics class shortcut" section has been added to the JavaTemplate documentation.

## What's your motivation?
The Semantics class has already been used in some of our recipes (like [IsNotEmptyToJdk](https://github.com/openrewrite/rewrite-apache/blob/d83e6f47b240ae9959684c5927861db91e19157c/src/main/java/org/openrewrite/apache/commons/lang/IsNotEmptyToJdk.java#L74-L77)). No documentation was available about this worked exactly.

### DIFF
--- a/docs/concepts-and-explanations/javatemplate.md
+++ b/docs/concepts-and-explanations/javatemplate.md
@@ -22,7 +22,8 @@ public class ChangeMethodInvocation extends JavaIsoVisitor<ExecutionContext> {
                 JavaParser.fromJavaVersion()                                   // Parser
                     .classpath("example-utils"))                               // Classpath lookup
             .staticImports("org.example.StringUtils.withString")               // Additional import
-            .doBeforeParseTemplate((String template) -> {})                    // Optional side-effect
+            .doAfterVariableSubstitution(System.out::println)                  // Optional side-effect
+            .doBeforeParseTemplate(System.out::println)                        // Optional side-effect
             .build();
 }
 ```


docs/concepts-and-explanations/javatemplate.md
```diff
@@ -274,4 +275,39 @@ When a JavaTemplate is used for matching, then the `Matcher` can be used to "ext
...
+        Semantics.statement(this, "Example", () -> { String example = "some statement"; }).build();
+```
+
+Behind the scenes, the annotation processor transforms these `Semantics.expression(...)` and `Semantics.statement(...)` calls into actual JavaTemplate definitions with proper substitution indicators and imports.
+
+:::info
+Prerequisite: To use this feature, make sure your project is set up for Refaster-style recipes with annotation processing enabled. See the [How to create a Refaster recipe](../authoring-recipes/refaster-recipes#how-to-create-a-refaster-recipe) section for more details.
+:::
+
+This pattern is especially useful when you're prototyping or when your templates are derived from simple expressions or statements that can be represented as lambdas. It's functionally equivalent to defining a JavaTemplate manually but with less boilerplate.
```
timtebeek, line 313: I only ever recall seeing `Semantics` used to match constructs where it would be awkward to match the individual LST elements, but I don't see that reflected in the docs here. I'd think we want to at least make that intended usage clear.

Not sure if `Semantics` has been tried as the source of a JavaTemplate that you _apply_, as opposed to _match_. 🤔 For the moment I have a bit of trouble thinking of a direct use case where Refaster would not be better used. Perhaps another reason to document the intended use.

docs/concepts-and-explanations/javatemplate.md
```diff
@@ -17,11 +17,12 @@ More advanced refactoring recipes often require the construction of complex [Los
...
-        JavaTemplate.builder("withString(#{any(java.lang.String)}).length()")   // Code Snippet
+        JavaTemplate.builder("withString(#{any(java.lang.String)}).length()")  // Code Snippet
             .javaParser(
-                JavaParser.fromJavaVersion()                                    // Parser
-                    .classpath("example-utils"))                                // Classpath lookup
-            .staticImports("org.example.StringUtils.withString")                // Additional import
+                JavaParser.fromJavaVersion()                                   // Parser
+                    .classpath("example-utils"))                               // Classpath lookup
+            .staticImports("org.example.StringUtils.withString")               // Additional import
+            .doBeforeParseTemplate((String template) -> {})                    // Optional side-effect
```
Laurens-W, line 25: Would it be good to outline both `doAfterVariableSubstitution` and `doBeforeParseTemplate` here? With an example that just prints them, something like `System.out::println`

jevanlingen: Yes: https://github.com/openrewrite/rewrite-docs/pull/399/commits/2d70bd5fe184846ca7eef8b5c8ba4cfe3007ea76!

PR #401, opened 2025-07-16
Document single-recipe option passing for Maven plugin without full build

jevanlingen:
## What's changed?
Since https://github.com/openrewrite/rewrite-maven-plugin/pull/816 it is possible to apply options  when running a single recipe with maven. The docs did state this in the FAQ, but not on the "Running Rewrite on a Maven project without modifying the build" page. 

## What's your motivation?
Updating old docs with current features.


mike-solomon (approved):
Thanks!

PR #406, opened 2025-07-28
Fix version selector documentation

MrThaler:
## What's changed?

Changed the documentation of the XRange selector to represent the actual behavior.

Also added a missing comma in another selector.

## What's your motivation?

- Closes https://github.com/openrewrite/rewrite/issues/5779

There was a brief consideration to adjust the implementation to confirm to the documentation instead. However, the code and tests made it very clear that the observed behavior was intentional and that a "fix" might've introduced a lot of unexpected behavior.


timtebeek (approved):
Perfect, thanks a lot for bringing clarity here!

PR #432, opened 2025-10-23
Add type attribution reference

sambsnyd:
Perhaps we want to also add a shorter "type attribution" entry under concepts & explanations which could stick to describing the nature and importance of type attribution in a more general way.

### DIFF
--- a/docs/reference/type-attribution.md
+++ b/docs/reference/type-attribution.md
@@ -1,88 +1,90 @@
-# Type Attribution
+---
+description: A detailed, technical look into type attribution and how OpenRewrite uses it.
+---
 
-Type attribution is the process of associating semantic type information with syntactic elements in the Lossless Semantic Tree (LST).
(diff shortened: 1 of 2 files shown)


docs/reference/type-attribution.md
```diff
@@ -0,0 +1,385 @@
...
+    * Type parameters declared by the method
+
+* **`JavaType.Variable`** - Represents a field or variable:
+    * Owner (the class or method declaring the variable)
+    * Variable type
+    * Variable name
+
+* **`JavaType.Unknown`** - Singleton sentinel value for unresolved types
+
+* **`JavaType.Intersection`** - Represents intersection types (e.g., `T extends A & B`)
```
knutwannheden: Minor correction: The intersection type is currently only used when doing casts like `(Function & Serializable)`, which is a really obscure Java feature.

PR #436, opened 2025-11-07
JavaScript recipe authoring skill for Claude

knutwannheden:
This skill provides comprehensive guidance for creating OpenRewrite recipes in TypeScript to transform JavaScript/TypeScript codebases.

**Included files:**
- `SKILL.md` - Main skill documentation with quick reference
- `references/examples.md` - Complete recipe examples
- `references/patterns-and-templates.md` - Pattern matching and template system guide
- `references/testing-recipes.md` - Comprehensive testing guide

**Key features:**
- Pattern matching and templating system
- Visitor-based transformations
- Recipe configuration with `@Option` decorators
- Scanning recipes for multi-pass analysis
- Comprehensive testing strategies with `afterRecipe` and `beforeRecipe`
- Type-safe AST traversal patterns

All examples use the published `@openrewrite/rewrite` NPM package with correct API signatures.

### DIFF
--- a/openrewrite-recipe-writer/skills/writing-openrewrite-recipes-js/references/lst-concepts.md
+++ b/openrewrite-recipe-writer/skills/writing-openrewrite-recipes-js/references/lst-concepts.md
@@ -848,7 +848,7 @@ const tmpl = template`isDate(${capture('value')})`
         }
     });
 
-return await tmpl.apply(this.cursor, method, match);
+return await tmpl.apply(this.cursor, method, {values: match});
 ```
 
 **When to use templates:**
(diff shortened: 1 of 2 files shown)


bryanfriedman:
This all LGTM. Two questions just for consideration:

- Maybe it makes more sense to have the skills be separate as you've done since they are so different. But I'm just wondering if it's worth trying to combine into one skill that can take either path depending on what you ask for? (Not sure if this will overly complicate it though, so fine leaving as is for now, but just a thought.)
- If we do keep them as separate skills:
  - Do we want to add a command for `create-recipe-js` like we have for `create-recipe`?
  - Do we want to rename the existing skill as `writing-openrewrite-recipes-java`?

/cc @sjungling

PR #450, opened 2025-12-31
Migrate from Immer to Mutative for immutable LST updates

knutwannheden:
- Replace `immer` with `mutative` dependency in setup docs and npm install commands
- Change all `produce()` calls to `create()` from mutative across examples and templates
- Update recipe registration to use `RecipeMarketplace` with `CategoryDescriptor` category paths instead of `RecipeRegistry`
- Add documentation for category hierarchy patterns with nested descriptors
- Maintain `produceAsync()` from `@openrewrite/rewrite` for async operations


mike-solomon (approved):
Thanks for updating this!

~I'll create an issue to update the JS docs to reflect these changes.~ whoops missed the fact that you already did this.

Issue #492, opened 2026-05-01
Visual refresh: typography, cards, sidebar, admonitions


LindsayAuch:
## Summary

A multi-part design refresh of the OpenRewrite docs to align styling with a simpler, Inter-typography card-based aesthetic.

Tracked as 5 draft PRs against an integration branch (`LindsayAuch/visual-refresh-sandbox`) so each slice gets its own diff while staying off `master` until reviewed as a whole. A sandbox docs page (`/sandbox`) renders one example of every new pattern for visual review.

## Scope

- **Typography** — drop Beausite, Inter throughout, smaller H1, weight/size tweaks
- **Card components** — full-width stacked cards, smaller padding/height, light-grey hover, swizzled DocCard cleanup, gems-bg removal
- **Sidebar nav** — bold categories, blue active states, tighter spacing
- **Markdown link-list cards** — recipe README pages render link-only lists as cards
- **Admonitions** — Info / Tip / Note / Warning / Caution / Danger using Lucide outlined icons + per-variant color schemes

## Process

- Foundation PR (sandbox page + integration branch) lands first
- 5 styling PRs target the sandbox branch, not `master`
- After approval, the sandbox branch becomes a single PR into `master`

## Backup

Full WIP branch: [`LindsayAuch/live-local-preview`](https://github.com/openrewrite/rewrite-docs/tree/LindsayAuch/live-local-preview)

LindsayAuch:
All draft PRs opened. Targets:

- **Integration:** #493 — `LindsayAuch/visual-refresh-sandbox` → `master` (do not merge until children land)
- **Children** (target the integration branch, not master):
  - #494 — Typography refresh
  - #495 — Card components redesign
  - #496 — Sidebar nav refresh
  - #497 — Markdown link-list cards
  - #498 — Admonition Lucide icons

Sandbox preview lives at `/sandbox` on the integration branch. Each child PR shows its diff in isolation; the cumulative effect renders once they're all merged into the sandbox branch.

Backup of all changes pre-applied: [`LindsayAuch/live-local-preview`](https://github.com/openrewrite/rewrite-docs/tree/LindsayAuch/live-local-preview).

PR #493, opened 2026-05-01
Visual refresh of docs site 

LindsayAuch:
- Integration branch for the visual refresh effort tracked in #492.

Five draft PRs target this branch. **Do not merge until all child PRs land here and the result is reviewed as a whole.**

## Sandbox preview

- `/sandbox` page lives in this branch — renders one example of every new pattern.
- Each child PR applies its slice on top of this branch; preview deployment shows the cumulative effect.

## Child PRs

| # | Scope |
|---|-------|
| 1 | Typography refresh |
| 2 | Card components redesign |
| 3 | Sidebar nav refinement |
| 4 | Markdown link-list cards |
| 5 | Admonition Lucide icons |

## Backup

Full WIP branch with everything pre-applied: [`LindsayAuch/live-local-preview`](https://github.com/openrewrite/rewrite-docs/tree/LindsayAuch/live-local-preview).

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -436,3 +436,37 @@ html[data-theme='dark'] {
   margin: 0 calc(-1 * var(--ifm-pre-padding));
   padding: 0 calc(var(--ifm-pre-padding) - 3px) 0 var(--ifm-pre-padding);
 }
+
+/* Style markdown link-only lists as cards (recipe README pages) */
+.markdown ul:has(> li > a:only-child) {
+  list-style: none;
+  padding-left: 0;
+}
+
+.markdown ul:has(> li > a:only-child) > li {
+  margin: 0 0 0.5rem 0;
+  padding: 0;
+}
+
+.markdown ul:has(> li > a:only-child) > li::before {
+  content: none;
+}
+
+.markdown ul:has(> li > a:only-child) > li > a {
+  display: block;
+  padding: 0.75rem 1.25rem;
+  border: 1px solid var(--ifm-color-border);
+  border-radius: 8px;
+  background-color: #fff;
+  color: var(--ifm-heading-color);
+  font-weight: 600;
+  font-size: 1.0625rem;
+  text-decoration: none;
+  transition: background-color 0.15s ease, border-color 0.15s ease;
+}
+
+.markdown ul:has(> li > a:only-child) > li > a:hover {
+  background-color: #F3F4F6;
+  border-color: rgba(4, 24, 52, 0.2);
+  text-decoration: none;
+}
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -436,3 +436,69 @@ html[data-theme='dark'] {
   margin: 0 calc(-1 * var(--ifm-pre-padding));
   padding: 0 calc(var(--ifm-pre-padding) - 3px) 0 var(--ifm-pre-padding);
 }
+
+/* ============================================================
+ * Visual polish — neo card design tokens, rounded media,
+ * sidebar active-leaf weight bump.
+ * Additive overrides; safe to land regardless of which other
+ * visual-refresh PRs merge first.
+ * ============================================================ */
+
+:root {
+  --neo-border-card: rgba(4, 24, 52, 0.12);
+  --neo-grey-300: rgba(4, 24, 52, 0.25);
+  --neo-shadow-card: 0 1px 2px rgba(4, 24, 52, 0.06), 0 4px 12px rgba(4, 24, 52, 0.05);
+  --neo-card-hover-bg: #FFFFFF;
+}
+
+html[data-theme='dark'] {
+  --neo-border-card: rgba(255, 255, 255, 0.1);
+  --neo-grey-300: rgba(255, 255, 255, 0.25);
+  --neo-shadow-card: 0 1px 2px rgba(0, 0, 0, 0.3), 0 4px 12px rgba(0, 0, 0, 0.2);
+  --neo-card-hover-bg: rgba(255, 255, 255, 0.04);
+}
+
+/* DocCard — transparent default, white on hover with shadow lift */
+[class^=cardContainer_] {
+  background-color: transparent !important;
+  border-color: var(--neo-border-card) !important;
+  transition: background-color 120ms ease, border-color 120ms ease, box-shadow 120ms ease !important;
+}
+
+[class^=cardContainer_]:hover {
+  background-color: var(--neo-card-hover-bg) !important;
+  border-color: var(--neo-grey-300) !important;
+  box-shadow: var(--neo-shadow-card) !important;
+}
+
+/* Markdown link-list cards — same treatment */
+.markdown ul:has(> li > a:only-child) > li > a {
+  background-color: transparent !important;
+  border-color: var(--neo-border-card) !important;
+  transition: background-color 120ms ease, border-color 120ms ease, box-shadow 120ms ease !important;
+}
+
+.markdown ul:has(> li > a:only-child) > li > a:hover {
+  background-color: var(--neo-card-hover-bg) !important;
+  border-color: var(--neo-grey-300) !important;
+  box-shadow: var(--neo-shadow-card) !important;
+}
+
+/* Rounded corners on inline media */
+.markdown img,
+figure img {
+  border-radius: 12px;
+}
+
+.markdown iframe[src*="youtube"],
+.markdown iframe[src*="youtu.be"],
+.markdown iframe[src*="vimeo"],
+.markdown iframe {
+  border-radius: 12px;
+}
+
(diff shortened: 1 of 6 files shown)


src/css/custom.css
```diff
@@ -202,10 +289,10 @@ html[data-theme=dark] [class^=docMainContainer_] .breadcrumbs__link:hover {
 }
 
 [class^=docMainContainer_] h1 {
-  --ifm-heading-font-weight: 300;
-  --ifm-heading-font-family: "Beausite", sans-serif;
-  --ifm-h1-font-size: 3.75rem;
-  --ifm-leading: 1;
+  --ifm-heading-font-weight: 600;
+  --ifm-heading-font-family: "Inter", sans-serif;
```
mike-solomon, line 298: I see we're switching from Beausite to Inter in a couple places - but at the top of this file we still declare the font-family of Beausite (line 8). I suspect we want to remove that at the top of this file - unless we want to use Beausite in other places?

src/theme/DocCategoryGeneratedIndexPage/styles.module.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.tocLabel::before {
+  content: '';
+  width: 16px;
+  height: 16px;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' fill='none' stroke='%23041834' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'%3E%3Cline x1='3' y1='6' x2='21' y2='6'/%3E%3Cline x1='3' y1='12' x2='15' y2='12'/%3E%3Cline x1='3' y1='18' x2='18' y2='18'/%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-size: contain;
+}
```
mike-solomon, line 43: This doesn't work in dark mode: 

Image

Image

src/css/custom.css
```diff
@@ -82,11 +89,15 @@ html[data-theme='dark'] {
...
   --ifm-heading-line-height: 1.25;
-  .theme-code-block {
-    border: 1.5px solid var(--ifm-color-border);
-  }
+
+  --neo-border-card: rgba(255, 255, 255, 0.14);
+  --neo-grey-300: rgba(255, 255, 255, 0.3);
+  --neo-shadow-card: 0 1px 2px rgba(0, 0, 0, 0.3), 0 4px 12px rgba(0, 0, 0, 0.25);
+  --neo-card-hover-bg: rgba(255, 255, 255, 0.06);
+  --neo-card-bg: rgba(255, 255, 255, 0.025);
```
mike-solomon, line 90: (`neo-card-bg`) This appears to only be declared in dark mode.

src/css/custom.css
```diff
@@ -82,11 +89,15 @@ html[data-theme='dark'] {
   --ifm-link-color: #85fe99;
   --ifm-toc-border-color: transparent;
   --ifm-section-divider-color: #d9d9d9;
-  --ifm-card-background-color: #fff;
   --ifm-heading-line-height: 1.25;
-  .theme-code-block {
```
mike-solomon, line 87: We lost the border around code blocks in dark mode. I feel like this makes it a bit harder to read.

Old:

Image

New: 

Image

mike-solomon (changes requested):
I'm not super experienced with CSS or TypeScript - so hoping @sjungling could help out with that front. I do see a few issues that I pointed out below, though.

src/css/custom.css
LindsayAuch: Beausite is great for the product marketing site but, for a technical document I have not seen a display font used in this manner especially with very long titles."Running Rewrite on a Maven project without modifying the build" is practically a sentence. Inter is already in use for body copy here, so keeping it for headings creates consistency and reduces the visual weight of these longer titles considerably.

mike-solomon: I think there may have been a misunderstand here - so let me try to rephrase.

Line 8 of this file still declares `font-family: Beausite`.

In this PR, you were removing Beausite in favor of Inter - but did not touch that line at the top of the file. 

The question is whether that was intentional (i.e., we intend to use Beausite in some places in the site) - or whether that was an accident and something you'll want to fix.

src/theme/DocCategoryGeneratedIndexPage/styles.module.css
LindsayAuch: Added a dark-mode override using `:global(html[data-theme='dark'])` so the menu-icon SVG swaps to white stroke (`#ffffff`) when dark mode is active.

src/css/custom.css
LindsayAuch: In light mode, the hover style transition from the stroked outline/no background state to a white background provides enough contrast on its own to clearly communicate the hover state change.That approach wasn’t achievable in dark mode (outline to just background) because the surrounding contrast is already much darker, so I added the stroke treatment in dark mode to create a clearer hover distinction and improve visibility. it looks kind of busy in light mode so I removed the stroke on the card hover.

src/css/custom.css
LindsayAuch: Code block in dark mode now has the subtle border

src/theme/DocCategoryGeneratedIndexPage/styles.module.css
LindsayAuch: Added a dark-mode override using :global(html[data-theme='dark']) so the menu-icon SVG swaps to white stroke (#ffffff) when dark mode is active.

src/css/custom.css
mike-solomon: I don't see any changes to the border. Did those changes not get checked in? Or were you saying that we intentionally don't want the border?

src/css/custom.css
LindsayAuch: yes, I misinterpreted the feedback. Removed the font declaration.

mike-solomon:
@LindsayAuch the code blocks now look quite wonky. There's too much padding and the font is smaller - making it difficult to read. 

Current site:

Screen Shot 2026-05-14 at 11 03 39 AM

Changes proposed in this PR:

Screen Shot 2026-05-14 at 11 03 58 AM

src/css/custom.css
LindsayAuch: Screenshot 2026-05-22 at 3 32 30 PM
 Looks like this on my machine and I asked claude to check in all of the changes

LindsayAuch:
All three concerns addressed:

Code font: bumped --ifm-code-font-size 81.25% → 87.5% (13px → 14px).
Dark-mode border: doubled alpha 0.12 → 0.25, visible against the code-block background.
Padding: reduced inner vertical padding to 0.5rem via a scoped .theme-code-block code rule. First attempt used --ifm-pre-padding but that variable is also read by card CSS modules — a global change rippled into cards on generated-index pages. The scoped rule keeps the fix local to code blocks.

Screenshots: 
Screenshot 2026-05-22 at 5 15 45 PM

Screenshot 2026-05-22 at 5 19 04 PM

Screenshot 2026-05-22 at 5 19 15 PM
c07-4498-817b-adb7b4dbfa09" />



Re-requesting your review.

src/css/custom.css
LindsayAuch: Sorry for the confusion before — diagnosed the real issue today. The border was checked in, but at rgba(255, 255, 255, 0.12) it was effectively invisible against the code-block background, which is why neither of us could see it on rebuild.
Bumped the alpha to 0.25 in this PR. Dark-mode code block now:
Screenshot 2026-05-22 at 5 19 04 PM

Resolving this thread — if it still doesn't read clearly on your machine, reopen and I'll bump alpha further.

mike-solomon (changes requested):
I couldn't quite tell what you were saying in your previous reply as it was improperly formatted and contained lots of random magic numbers.

That being said, there are quite a few issues I'm seeing still. To list some that immediately jump out to me:

1. We need titles for some of the admonitions. It's fine to get rid of "info" or "tip" as a title - but there are many cards that provide information in the title field. Additionally, admonitions aren't accessible. Removing the titles from them is ok if we pair it with adding an aria-label. You can direct Claude to the Moderne Docs for how to do this correctly. 
2. The spacing of text is very wonky. It would be a good idea to run your branch locally and open it compared to the current OR docs so you can see the changes. The sidebar is too smushed but things around bullet points have way too much padding for instance.
3. The admonitions don't match the Moderne Docs. Unsure if that's intentional or not - but either way they are too muted in color scheme.

Here's a photo with some of the things pointed out:

Image

Current docs for comparison:

Image

mike-solomon:
With the plan for redesigning the sites in a different way, I feel like this PR no longer makes sense so going to close it. Please let me know if you disagree.

PR #522, opened 2026-07-28
Maven Central -> Code Genome Project

mike-solomon:


docs/authoring-recipes/data-tables.md
timtebeek, line 321: Should we add the same to the Maven tab here?

docs/authoring-recipes/recipe-development-environment.md
timtebeek, line 441: I think we'd best add the Maven repository equivalent here as well.

timtebeek (approved):

